PR #91, opened 2014-10-31
Force pre-2.2 behavior for circular variable.

phyllisstein:
Trunk currently emits a warning about the circular argument reference [here](https://github.com/fog/fog-core/blob/master/lib/fog/core/collection.rb#L144). If I'm reading [the discussion](https://bugs.ruby-lang.org/issues/10314), adding parenthesis to `filters` should force the old behavior and suppress the warning. Not much of a PR, but it's working great locally and didn't break the test suite.


tokengeek:
Closing since #100 has been merged, renaming and making this change redundant.

Thanks!

PR #113, opened 2014-12-21
[GH105] Extract service module code - WIP

tokengeek:
**Still a work in progress - Please do not merge**

This is another step in moving towards a cleaner provider/service registry.

This uses general conventions that most providers adhere to and some
workarounds to attempt to require either `fog/provider/service` or `fog/service/provider`
then initialise `Fog::Service::Provider` or `Fog::Provider::Service`

So this eliminates a lot of the provider specific stuff behind the abstraction code for
`Fog::Compute[:provider]`

There is unfortunately still a little lookup table for providers that reformatted their names so
`:providername` != `require "fog/provider_name"` != `class Providername`

Additional overloading has also been lost here, notably `:version` options.

There is also an underlying implementation issue where the current default of
`Fog::Service::Provider` works in most cases except a few newer extractions and the
fallback works. The fallback is the preferred form `Fog::Provider::Service`.

Unfortunately `fog-brightbox` and `fog-softlayer` have implemented `Fog::Provider::Service`
as modules for support classes and so do not respond to `new`

So this triggers in the current (legacy) non-preferred order until those providers
are fixed.


plribeiro3000:
@tokengeek As far as i can tell this is ready. Am i wrong?

Do you need an extra hand to add some tests around this?

tokengeek:
@plribeiro3000 - This isn't 100% compatible because some providers have added `version` arguments at this level, namely HP and Rackspace.

So merging this will break core for anyone relying on `:version`.

Brilliantly they have opposite defaults so with HP the default is V1 whilst Rackspace defaults to V2 so you can't just remove it since it'll affect someone in some combination.

BOTH providers need to be updated, or I implement some intermediate hack which I haven't thought out fully.

There's also so much other churn on `fog` I'm barely keeping track of all of the changes and didn't want to try anything too experimental with the number of issues that seem to keep surfacing.

cphrmky:
I've opened #130, a PR into this branch (105_extract_service_module_code) that reduces duplicated code.

plribeiro3000:
Half of the intention on this PR was made on #156 which is already merged.

So i believe we need to rebase this branch or start over. Either way i will close this PR and create an issue to keep tracking of this.

PR #114, opened 2014-12-23
Rescue exception with the console

### LINKED ISSUE #112: Fog::Logger write occasionally raises Errno::EBADF
https://github.com/fog/fog-core/blob/master/lib/fog/core/logger.rb#L38

Some small percentage of s3 operations would fail because a call to Fog:Logger would raise Errno::EBADF. (I am pretty sure the call to log is because of the path_style warning.) Note that most operations would still be fine even if they did invoke the warning; the exception wouldn't be raised and I could see the log entry. I have no way to reproduce. It's very possible it has to do with us capturing stdout and the log file getting rotated?

To handle it I simply added:

```
[:debug, :warning, :deprecation].each { |level| ::Fog::Logger[level] = nil }
```

to our initialization code. However, I feel an exception from logging shouldn't fail a real operation like uploading an object.

If you are against rescuing Errno::EBADF here then I can understand. Just trying to help others down the road.

lardcanoe:
fixes #112

### DIFF
--- a/lib/fog/core/logger.rb
+++ b/lib/fog/core/logger.rb
@@ -39,7 +39,8 @@ module Fog
       end
       nil
     rescue ::Errno::EBADF
-      # Prevent issues with the console from preventing the main operation from finishing
+      # Prevent issues with the console from affecting the main operation
+      # Such as uploading an object to S3
       return nil if @channels[key] == ::STDERR
       raise
     end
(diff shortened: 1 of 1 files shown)


plribeiro3000:
Closing by lack of feedback. Feel free to reopen or comment it to continue discussion.

geemus:
Thanks!

PR #115, opened 2014-12-23
Simpler check for Ruby version and skip

tokengeek:
This removes the need for Rubygems just to parse the `RUBY_VERSION`
string to ignore 1.8.7

String comparison gives us a simpler way to check this and I started
seeing errors about frozen Strings being modified related to this.

This now "skips" on any ruby < 1.9. string to ignore 1.8.7

String comparison gives us a simpler way to check this and I started
seeing errors about frozen Strings being modified related to this.

This now "skips" on any ruby < 1.9.3

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,5 +1,3 @@
-require "rubygems"
-
 require "minitest/autorun"
 require "minitest/spec"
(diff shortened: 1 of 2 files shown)


geemus:
Thanks!

PR #116, opened 2015-01-03
Use Fog::Formatador

starbelly:
Prefix `Formatador.new` call  with scope resolution op so we don't call `Fog::Formatador.new`

### DIFF
--- a/lib/fog/core/model.rb
+++ b/lib/fog/core/model.rb
@@ -21,16 +21,7 @@ module Fog
     end
 
     def inspect
-      Thread.current[:formatador] ||= Formatador.new
-      data = "#{Thread.current[:formatador].indentation}<#{self.class.name}"
-      Thread.current[:formatador].indent do
-        unless self.class.attributes.empty?
-          data << "\n#{Thread.current[:formatador].indentation}"
-          data << self.class.attributes.map { |attribute| "#{attribute}=#{send(attribute).inspect}" }.join(",\n#{Thread.current[:formatador].indentation}")
-        end
-      end
-      data << "\n#{Thread.current[:formatador].indentation}>"
-      data
+      Fog::Formatador.format(self)
     end
 
     def reload
--- a/lib/fog/core/collection.rb
+++ b/lib/fog/core/collection.rb
@@ -70,7 +70,7 @@ module Fog
     end
 
     def inspect
-      Fog::Formatador.new(self)
+      Fog::Formatador.format(self)
     end
 
     def load(objects)
--- a/lib/fog/core/ssh.rb
+++ b/lib/fog/core/ssh.rb
@@ -109,14 +109,14 @@ module Fog
       def display_stdout
         data = stdout.split("\r\n")
         if data.is_a?(String)
-          Formatador.display_line(data)
+          Fog::Formatador.display_line(data)
         elsif data.is_a?(Array)
-          Formatador.display_lines(data)
+          Fog::Formatador.display_lines(data)
         end
       end
 
       def display_stderr
-        Formatador.display_line(stderr.split("\r\n"))
+        Fog::Formatador.display_line(stderr.split("\r\n"))
       end
 
       def initialize(command)
--- a/lib/fog/formatador.rb
+++ b/lib/fog/formatador.rb
@@ -1,65 +1,71 @@
 module Fog
   # Fog::Formatador
-  class Formatador
-    attr_accessor :object, :string, :formatador
-
-    def initialize(obj)
-      @object = obj
-      @formatador ||= ::Formatador.new
+  module Formatador
+    def self.formatador
+      Thread.current[:formatador] ||= ::Formatador.new
     end
 
-    def to_s
-      return string unless string.nil?
-      init_string
-      indent { string << object_string }
-      (string << "#{indentation}>").dup
+    def self.format(object, opts = { :include_nested => true })
+      string = init_string(object)
+      indent { string << object_string(object, opts) }
+      string << "#{indentation}>"
+    end
+
+    def self.display_line(data)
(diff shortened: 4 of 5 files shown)


plribeiro3000:
@starbelly can't we make `Fog::Formtador` reusable and use it here as well?

starbelly:
Definitely :)

lib/fog/core/model.rb
```diff
@@ -21,16 +21,7 @@ def initialize(new_attributes = {})
...
-      data = "#{Thread.current[:formatador].indentation}<#{self.class.name}"
-      Thread.current[:formatador].indent do
-        unless self.class.attributes.empty?
-          data << "\n#{Thread.current[:formatador].indentation}"
-          data << self.class.attributes.map { |attribute| "#{attribute}=#{send(attribute).inspect}" }.join(",\n#{Thread.current[:formatador].indentation}")
-        end
-      end
-      data << "\n#{Thread.current[:formatador].indentation}>"
-      data
+      Fog::Formatador.new(self)
```
geemus: This part confuses me. I can see using a Fog::Formatador instead, but doesn't the other logic here still need to be around (instead of just initializing one and not calling anything with/on it?).

starbelly: The same logic is in Fog::Formatador ( comprised of init_string(), attribute_string(), and object_attributes() ), so we're definitely being DRY by taking out the logic in model.rb. But it also goes through nested objects, which is why I was asking if a parameter to prevent that from happening in this case. It seemed clear to me that yes a param is needed to prevent nested objects from being included in the string for this case, but I figured I would see what everyone thought first. As for the param, we could say ```Fog::Formatador.new(obj, level)``` where level is an integer, though that's kind of cryptic.

geemus: hmm. I guess I just expected a different interface. ie something like `Fog::Formatador.new(...).inspect(self)` or `Fog::Formatador.inspect(self)`. It isn't readily apparent to me, I guess, that initializing an object would do the same thing as the above code used to.

starbelly: I agree that the use of it is cryptic, you actually have to dig into the code to figure out what it's doing and that's not a good thing(tm).  I think what you're suggesting is probably the better way to go. I think the idea was to simply move the logic in its own container seeing as how it was taking up so much space in collection.rb and model.rb and to keep things DRY. This all began with trying to make rubocop happy, and I still think it should be appeased. So if we say Fog::Formatador.inspect(self), what about the issue of not including nested objects in the case of model.rb? What do you think the call to inspect should look like in that case? Also, the reason I made it a class and not a module was something you said about indentation getting whacky due to threads, iirc. I am much more in favor of having a module, this isn't anything that needs to keep state IMHO.

geemus: Yeah. I think I would be fine with a module. There are potential indentation + threads issues, but it seems unlikely we would run in to them here. Maybe a second argument for nested stuff, but perhaps as a hash of options, like `Fog::Formatador.inspect(self, :include_nested => false)` or something like that. Buys us a bit more flexibility as I suspect that might not be the only option we would think of as we centralize more stuff.

starbelly: I concur and I'll push that in a minute. Are you in favor of having a single formatador instance for the module and just re-initializing at the end of the inspect method or do you think a new formatador instance should be created per call?

plribeiro3000: I think that `Fog::Formatador.format(self, :include_nested => false) would fit better IMHO.

geemus: Or `display`, format doesn't imply stdout to me per se. Open to suggestions.

I'd probably just create a new formatador instance per call. Again, the thread safety issue probably won't come up, but this will limit the window of possible issue.

plribeiro3000: @geemus `display` kind of implies it is gonna just print something, but the class is gonna do more then just print. What about `pretty_print` or something like that?

plribeiro3000: In a second thought, i think that `display` might be enough. It is gonna to receive an object and display it somehow. :+1: for `display`.

plribeiro3000:
@starbelly :+1: for your work!

lib/fog/core/model.rb
starbelly: I think format might be better because the string isn't automatically displayed? I'll go ahead and push what I got then we can figure it out from there :)

starbelly:
By the way, do you guys have an irc channel or something similar, maybe jabber?

lib/fog/core/model.rb
geemus: OH! Then yeah, probably format. Clearly I am missing contexts.

geemus:
@plribeiro3000 yeah, one of those might be nice. It comes up infrequently enough that I tend not to worry about it too much. I also kind of like the asynchrony of emails/issues/etc, but I understand it slows things down at times.

plribeiro3000:
@geemus Agreed on the asynchronous point. Perhaps it might let it as is for now.

We can revisit this idea once again in the future if needed. :+1:

plribeiro3000:
@starbelly Can you rebase this please?

starbelly:
It's already been rebased, about 6 or 7 commits squashed into one. Are you not seeing it that way?

plribeiro3000:
Nope. I guess the merge of #120 just breaked it again.

starbelly:
Hmm, just did a merge and fixed the conflict. Should I still need to rebase?

geemus:
Seems fine, thanks!

plribeiro3000:
@geemus way faster then me on this one. Merge?

plribeiro3000:
haha. Twice in a row. I got be faster next time. =)

PR #117, opened 2015-01-06
Add OpenStack Ironic service

Ladas:
Adding OpenStack Ironic service for managing baremetals.

### DIFF
--- a/lib/fog/baremetal.rb
+++ b/lib/fog/baremetal.rb
@@ -1,0 +1,23 @@
+module Fog
+  module Baremetal
+    def self.[](provider)
+      new(:provider => provider)
+    end
+
+    def self.new(attributes)
+      attributes = attributes.dup # Prevent delete from having side effects
+      provider = attributes.delete(:provider).to_s.downcase.to_sym
+
+      if providers.include?(provider)
+        require "fog/#{provider}/baremetal"
+        return Fog::Baremetal.const_get(Fog.providers[provider]).new(attributes)
+      end
+
+      raise ArgumentError, "#{provider} has no baremetal service"
+    end
+
+    def self.providers
+      Fog.services[:baremetal]
+    end
+  end
+end
--- a/lib/fog/core.rb
+++ b/lib/fog/core.rb
@@ -52,6 +52,7 @@ require "fog/core/whitelist_keys"
 
 # service wrappers
 require "fog/account"
+require "fog/baremetal"
 require "fog/billing"
 require "fog/cdn"
 require "fog/compute"


plribeiro3000:
cc/ @tokengeek @geemus I guess we need 2 more provider needing this new namespace, right?

Is there another provider that could use such service?

geemus:
Yeah, perhaps.

@Ladas I'm not seeing related stuff in fog/fog at the moment, am I missing something? Just want to understand the context of how this fits in. Thanks!

Ladas:
I am working on it, not yet ready as pull request. I have the draft, now I need to test each method with real Ironic and write tests.

You can look at it here https://github.com/Ladas/fog/commit/171537dc24c77ab1ae6d683873c6aa8edbe18b16

, growing pretty big and that is just basic CRUD support :-)

If I understand the Shindo tests correctly, do we need some real Ironic endpoint in CI for that? Or it uses just mocks?

tokengeek:
@Ladas The CI needs to be testable with mocks. It's impossible to run all of the tests against all the live services.

`bundle exec rake travis` should work for anyone even if they don't have a rack full of hardware available!

tokengeek:
@plribeiro3000 I've not looked at this but we do need to be careful
about making the abstract version of `Baremetal` the Openstack one.

Very quick search and I can spot these other "bare metal" provisioners
with API support that `Baremetal` could abstract:

* Ubuntu MAAS - http://maas.ubuntu.com/
* Puppet's Razor - http://puppetlabs.com/solutions/next-generation-provisioning

There's others but some don't mention an API to talk to or just use PXE
and can't trigger hardware starts with IPMI (or rank lower than my attention span).

I've not used any of them so not sure what common behaviour they
have to reasonably abstract.

@Ladas That's not to say it's not useful, just it might be worth
us considering recommending `Fog::Openstack::Ironic.new()` for the
time being rather than creating `Baremetal` abstraction no one has
planned.

I'm not sure how much this touches on software defined networking with
it's ports and links stuff for example.

@geemus Thoughts?

Ladas:
@plribeiro3000 @geemus I've just made consistent naming with other services, so far it always matches the keystone service name. In OpenStack this service is called baremetal.
Then this piece of code is consistent https://github.com/Ladas/fog/blob/8fe81ad4debd93568d7dcdb2b63e44fae43f2655/lib/fog/bin/openstack.rb

I don't know the fog code that well, to know how to cleanly hack this piece of code around. :-)

Also with the code I am doing, I am not specifying any abstract interface of baremetal, the only abstract thing here is the baremetal name. Should there be abstract interface somewhere for this service? Looking at e.g. different cloud services, doesn't seem to me like there is some strict interface.

geemus:
baremetalcloud (already in fog) is another example (I think).

I think calling this baremetal is probably fine as long as we set it up just under openstack (a more abstract/general provider of this type should wait until we have more examples/info I think).

Hope that helps, certainly happy to discuss more though.

@Ladas got it, just let me know if you have further questions or need help on the draft. Since I hadn't seen it previously I just wasn't sure how the above stuff fit in to the picture. Thanks!

ohadlevy:
I might add foreman discovery is a bare metal provider (with an api of course) see http://theforeman.org and https://github.com/theforeman/foreman_discovery

tokengeek:
@Ladas In an ideal world there should be an abstract version that behaves the
same no matter the provider but (as is obvious) we've know thought about this
yet. 

Ideal world `Fog::Openstack::Ironic` is used so that the Openstack code is namespaced
and that implements the same interface as `Fog::Baremetal`

Most of the existing code is namespaced by service (e.g. `Fog::Compute::Provider`) 
which makes things very confusing.

But you're right in that there's not much in the way of standards for interfaces
which is why having two or three implementations to compare helps.

I'm behind on documenting all of this and getting concrete tests in place for
fog 2.0 where some of this stuff is clearer.

@ohadlevy Thanks. Seems to offer listing of "discovered_hosts" rather than
IPMI controlling "nodes" so serves as a useful example about name differences
and capabilities.

@geemus Yeah. The company name doesn't help clear things up! Looking at
their site they appear to do bare metal provisioning and virtual server provisioning
so I would argue the provider offers a `Compute` service and a `Baremetal` service.

Their API appears to support the bare metal service but their fog provider is
implemented as a `Compute` one.

I really should try to document in one place my thinking about the new stuff

Ladas:
@tokengeek yeah OpenStack had baremetal service under compute before (nova-baremetal), it got deprecated and Ironic is the baremetal service now.

@geemus @tokengeek So can this commit stay, or I need to do this https://github.com/Ladas/fog/blob/8fe81ad4debd93568d7dcdb2b63e44fae43f2655/lib/fog/bin/openstack.rb
different way?  I don't know what is the correct way to set it up just under OpenStack. :-) I am basically following the structure, that is already there and that requires this patch.

tokengeek:
I'll let @geemus decide.

There's an excellent chance the rest of the Openstack namespace may get switched in a later major release so this has the benefits of consistency with the rest.

geemus:
Yeah, it's tricky. I think naming it by what it does instead of what it is called (especially since there is precedent for it to change) is probably going to be better. Though it does then have the concern that we'll want to claim bare metal as an abstract case. I guess maybe we should do the right thing now and we can cross the abstract case bridge when we arrive there.

geemus:
@tokengeek naming aside, given your recent work, should this integrate with the service registry stuff somehow instead of being this way?

tokengeek:
@geemus Yeah. In theory `require "openstack"` will run something like `Fog::Core::Registry.register_service(:baremetal, Fog::Openstack::Ironic)` that would allow `Fog::Baremetal.new(provider: "openstack")` to return an instance of itself.

That's the medium term abstraction which is most compatible.

However returning an instance means `Fog::Baremetal` would have the collections like `nodes` which is not truly abstract yet. Of course we have that issue in many places already and will probably tackle Compute first.

Ladas:
@geemus @tokengeek So what is the outcome? Can it stay like it is with later re-factoring? Or I should rewrite it somehow with Fog::Core::Registry?

tokengeek:
Sorry for the confusion and the delays.

Basically `Fog::Core::Registry` doesn't yet exist so I'd say go with it as is.

I'm still leaving the final decision and merge to @geemus however.

geemus:
Yeah, I think as-is is good enough for now. I just was holding off on merging until the fog/fog portion was ready.

Ladas:
I will need this one merged and released, before the fog/fog part can be merged. Seems like release of fog-core could be released pretty soon?

geemus:
I had thought I would try to merge them at the same time, otherwise it seems likely we might deploy this part without the other, which seems undesirable. Does that make sense?

Ladas:
I think it will not pass CI. This needs to be first merged and released, the the fog/fog CI can pass.

There is CI error in fog/fog now because of missing Baremetal.

I'll make sure the other part will get in, we need it for ManageIQ. :-)

Ladas:
I suspect correctly that he fog/fog CI depends on released fog-core, right?

geemus:
Apologies for the delays, it has been a hectic couple weeks with some work travel and other things going on.

I know we kind of went round and round on this. I wonder if an in-between might not be best. ie both Fog::Openstack::Ironic and Fog::Baremetal have downsides. How about Fog::Openstack::Baremetal? I feel like that might simplify some things (I think it might remove the need for fog-core changes) and it could buy us time until we are ready to think about the right abstractions for Fog::Baremetal.

Does that make sense? What do people think?

cphrmky:
`Fog::Baremetal` would help with SoftLayer.  `fog-softlayer` has portions of code (mostly in `Fog::Compute::Softlayer::Server` model) that smell quite bad as they are coping with the fact that you can order both VMs and Baremetal but the APIs are different.

geemus:
@cphrmky - hmm. Good call. I guess I didn't want to hold up this merge on us figuring out the shared portions (though it sounds more and more like we should be talking about those shared portions also).

Ladas:
@geemus I would rather vote for consistency, which means Fog::Baremetal. For handling all Baremetal related stuff, this abstraction is solid, I think. Since it is not bringing any abstract interfaces or models, I don't see it as a problem.

Adding specific stuff to OpenStack could be done also, but it would not look that nice.

Please let me know what is the conclusion, I would really like to have basic Baremetal support in next release.

geemus:
@Ladas - I'm not sure I follow your argument. I would tend to agree that in time we would land on `Fog::Baremetal`, but in my mind that implies abstract interfaces/models (at least in so far as trying to standardize on collection/model names and also the attributes of them). My concern comes in releasing this at the abstract level and then having our hands tied about parts of the interface when we go to add more services. Does that help clarify? What do you think?

Ladas:
@geemus Right so, let's divide it in two parts. 1. How much does it tie our hands. 2. How does abstraction in Fog look like.

1. So this definitely brings:

1.a) Naming of models: node, chassis, driver and port. While you can definitely find synonyms for them, I think this set of names is one of the solid ones. And every Baremetal has to have those, so you can manipulate with it. So this shouldn't really tie our hands, we would have to decide on some names anyway. (we can also change it later while keeping backwards compatibility)

1.b) Naming of attributes. Again, you need the basic attributes that are there now in order to work with Baremetals. The naming should be even less important here, cause you can easily alias the attributes. And we can later rename the abstract set of attributes, while making it backwards compatible. 

2. The current state of abstract interfaces is kinda fuzzy. From what I see, some abstract models like compute, has bigger part standardized. It is also the only super model I found (Fog::Compute::Server). 

But the state of other varies a lot e.g. metering, is clearly OpenStack specific, while it contains also monitoring, maybe pieces of billing, there are separate concepts for that. 

Also implementation of others can differ a lot, e.g. network, where only the basic set is shared.

My conclusion
-------------------

All the above, plus the fact that OpenStack doesn't have any specific service, so all services are also abstract concepts, led me to creating Fog::Baremetal.

As @tokengeek wrote above "However returning an instance means Fog::Baremetal would have the collections like nodes which is not truly abstract yet. Of course we have that issue in many places already and will probably tackle Compute first."

Seems like we first need to create abstract classes, that would actually define interfaces, that each class has to implement. Since there are already models named differently, there should not be a problem to have also aliases for models to unify them, deprecating the old naming till the next big release. That would apply for Baremetal too, if we will need to rename Baremetal models.

Sorry for the long text. :-) I am fine with any implementation, so let me know your decision, so we can have Baremetal support in OpenStack soon, it's kinda blocking us more and more.

Thank you.

geemus:
Thanks for the detailed explanation. I think the backwards compatibility (or not) is probably the largest part of my concern. It sounds as though this shouldn't be too crazy though. I will go ahead and bring this in and we can continue to discuss/work from there. Sorry for blocking on this and thank you for your patience.

PR #125, opened 2015-01-17
Use Fog::Formatador

starbelly:
This takes care of any stragglers ( i.e, places where Formatador was being used directly )


geemus:
Thanks!

PR #130, opened 2015-02-06
D.R.Y. out extending and requiring with service wrappers.

cphrmky:


plribeiro3000:
@cphrmky Thanks! This looks a lot better. Just not sure if thats in the same path @tokengeek is working so i think he is the best to merge this.

cc/ @tokengeek

tokengeek:
I'll hold off until I decide on where the code is going.

Problem with this metaprogramming is that it these files may be the logical places to document how these services are used. Will see how it goes.

geemus:
Sounds good, thanks all, just let me know if there is anything I can help discuss.

plribeiro3000:
Since #113 was closed due a need for rebase i believe this also needs to be closed until its fixed.

PR #132, opened 2015-02-11
Allow apps to pass along own user agent

mdarby:
In relation to: https://github.com/fog/fog/issues/3433#issuecomment-73893922

### DIFF
--- a/lib/fog/core/connection.rb
+++ b/lib/fog/core/connection.rb
@@ -6,6 +6,26 @@ module Fog
     # modifications such as authentication or response object.
     #
     class Connection
+
+      class << self
+        @@user_agents = []
+
+        def add_user_agent(str)
+          if /\S+\/[\d|.]+/.match(str)
+            @@user_agents << str
+          else
+            raise "User Agent must be in <app name>/<app version> notation."
+          end
+        end
+
+        def user_agents
+          agents = @@user_agents.dup
+          agents << "fog/#{Fog::VERSION}" if defined?(Fog::VERSION)
+          agents << "fog-core/#{Fog::Core::VERSION}"
+          agents.uniq.compact.join(" ")
+        end
+      end
+
       # Prepares the connection and sets defaults for any future requests.
       #
       # @param [String] url The destination URL
@@ -34,10 +54,8 @@ module Fog
 
         params[:debug_response] = true unless params.key?(:debug_response)
         params[:headers] ||= {}
-        user_agent = "fog-core/#{Fog::Core::VERSION}"
-        user_agent = "fog/#{Fog::VERSION} #{user_agent}" if defined?(Fog::VERSION)
-        params[:headers]["User-Agent"] ||= user_agent
         params.merge!(:persistent => params.fetch(:persistent, persistent))
+        params[:headers]["User-Agent"] ||= user_agent
         @excon = Excon.new(url, params)
       end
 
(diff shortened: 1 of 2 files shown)


lib/fog/core/connection.rb
```diff
@@ -77,6 +75,13 @@ def reset
 
       private
 
+      def user_agent
+        arr = [ENV['FOG_CUSTOM_USER_AGENT']] # Allow apps to pass custom UA for metrics, etc.
```
geemus: This should probably also use an `if` instead of just adding nil (otherwise when you join you end up with leading whitespace, which is why the tests break).

geemus: and/or lstrip or something.

everett-toews:
Can this be documented in the source somewhere that users can become aware of it?

Any such documentation should also include an example of how to set a proper User-Agent as it's really easy to get wrong. e.g. `my-app/1.2`

mdarby:
Oh no doubts -- this was just to make sure we were on the same page.

lib/fog/core/connection.rb
mdarby: I added `compact`down here to take care of that.

fcheung:
Any reason to prefer an env variable over an option passed to `initialize` ?

mdarby:
Ease honestly.

This connection method is really quite buried and I wouldn't want to manually pass a bit of info through the layers of Fog. Seems to me that if someone wanted to use this feature they could more easily control it via an optional ENV var.

fcheung:
How about a class level property `Fog::Connection.user_agent = ...` ?

mdarby:
If you'd prefer that method I can easily refactor. :)

plribeiro3000:
I think that perhaps a variable through fog would better because the user would be able to use both options. He can define the code to search on `ENV` manually and thus have the same behavior. But thats just my opinion. =)

fcheung:
I feel a bit squiffy about using environment variables, unless the intention truly is that one should be able to change this without changing source code

tokengeek:
I think I'm with @fcheung on this. `ENV` are great for applications but `fog-core` and the majority of `fog` is intended for use as a library.

If people want to see that `Carrierwave` is making requests, `carrierwave` shouldn't be injecting ENVs into other people's applications.

That's my take on it anyway.

The majority of the `ENV` settings in `fog` are for using the binary but maybe some leaked through.

mdarby:
Thoughts?

1 irb ruby 2015-02-12 12-07-43

fcheung:
I think that if you do it like that subclasses (e.g. Fog::XML::Connection) won't inherit the value. Perhaps one of the few times where an @@var is appropriate

mdarby:
I've refactored to use `@@var`.
I added a `Fog::Connection.add_user_agent(str)` method that checks for format.

updated

mdarby:
Direction approval?

geemus:
Seems reasonable to me, I'll let the others add their 2 cents though. Also, I'm not altogether opposed to ENV also overriding this, I just agree that I would prefer for it to not be the ONLY way of setting this value. I think there can be some value to supporting both.

mdarby:
Bump

GTG for docs/tests?

plribeiro3000:
`Fog::Core::Connection.add_user_agent("my-app/1.2")` seem reasonable.

:+1: from me!

geemus:
Thanks!

geemus:
released in 1.29.0

PR #134, opened 2015-03-05
value is an arbitrary object and may not implement #== with string arg

fcheung:
On ruby 2.2 the changed line can emit a warning along the lines of

> Comparable#== will no more rescue exceptions of #<=> in the next release.

One case where this happens is when value is an instance of Time. I don't this this change affects the case we're trying to capture (at least not the one mentioned in the specs) but it avoids the warning.


geemus:
Thanks!

PR #135, opened 2015-03-14
Update Hash helper specs

tokengeek:
Replaces some old Shindo tests with MiniSpec versions.

Whilst working on this spotted that coverage was set up incorrectly
so under-reported.

When that worked there was a obviously unreachable section of code
I removed.

Note `shindo` is not a dependency but it "appears" due to some module
skulduggery in `spec/test_helpers/formats_helper_spec` I believe.


tokengeek:
Confirmed that there were 1.8.7 issues on `master` with `WhitelistKeys.whitelist` that were just untested.

geemus:
Thanks!

On Sat, Mar 14, 2015 at 7:51 PM, Paul Thornthwaite <notifications@github.com

PR #137, opened 2015-03-19
only check for `to_time` as a last resort.

tenderlove:
Some libraries (like Rails) will monkey patch String to add the
`to_time` method.  That means this method will function differently
depending on who monkey patched String first.  For example, [Rails changed it's default on `to_time`](https://github.com/rails/rails/commit/b79adc4323ff289aed3f5787fdfbb9542aa4f89f) to parse with regard to local time rather than UTC, which means that if you use fog with Rails, the dates returned will differ depending on which version of Rails you use.

If the object is a String, lets parse it so that the function has a
consistent return value for a given input.


plribeiro3000:
Thanks @tenderlove!

geemus:
Thanks!

tenderlove:
Thank you! <3<3<3

PR #139, opened 2015-03-23
[GH #138] Fix Formatador for objects defining `map`

### LINKED ISSUE #138: Formatador breaks with single objects with `#map` defined.
Brightbox's `CloudIp` object can no longer be inspected because `#map`
is already defined to "port" them between servers.

```
>> service.cloud_ips
ArgumentError: wrong number of arguments (0 for 1)
        from /home/paul/.gem/ruby/2.2.1/gems/fog-brightbox-0.7.1/lib/fog/brightbox/models/compute/cloud_ip.rb:32:in `map'
        from /home/paul/.gem/ruby/2.2.1/gems/fog-core-1.29.0/lib/fog/formatador.rb:87:in `inspect_object'
        from /home/paul/.gem/ruby/2.2.1/gems/fog-core-1.29.0/lib/fog/formatador.rb:74:in `block in nested_objects_string'
        from /home/paul/.gem/ruby/2.2.1/gems/formatador-0.2.5/lib/formatador.rb:92:in `indent'
        from /home/paul/.gem/ruby/2.2.1/gems/fog-core-1.29.0/lib/fog/formatador.rb:43:in `indent'
        from /home/paul/.gem/ruby/2.2.1/gems/fog-core-1.29.0/lib/fog/formatador.rb:74:in `nested_objects_string'
        from /home/paul/.gem/ruby/2.2.1/gems/fog-core-1.29.0/lib/fog/formatador.rb:56:in `object_string'
        from /home/paul/.gem/ruby/2.2.1/gems/fog-core-1.29.0/lib/fog/formatador.rb:14:in `block in format'
        from /home/paul/.gem/ruby/2.2.1/gems/formatador-0.2.5/lib/formatador.rb:92:in `indent'
        from /home/paul/.gem/ruby/2.2.1/gems/fog-core-1.29.0/lib/fog/formatador.rb:43:in `indent'
        from /home/paul/.gem/ruby/2.2.1/gems/fog-core-1.29.0/lib/fog/formatador.rb:14:in `format'
        from /home/paul/.gem/ruby/2.2.1/gems/fog-core-1.29.0/lib/fog/core/model.rb:24:in `inspect'
        from /home/paul/.gem/ruby/2.2.1/gems/fog-core-1.29.0/lib/fog/formatador.rb:87:in `block in inspect_object'
        from /home/paul/.gem/ruby/2.2.1/gems/fog-core-1.29.0/lib/fog/core/collection.rb:19:in `map'

```

It seems when the code that is now `Fog::Formatador` was extracted from
`CollectionString` it became more general rather than just for collections.

So when trying to enumerate the collection of Cloud IPs, because `map` is assumed rather than `each` (which appears to be an implementation detail here https://github.com/fog/fog-core/blob/master/lib/fog/formatador.rb#L87) an error occurs due to the arity difference.

I plan to deprecate `map` in `fog-brightbox` however this is still a bug in Formatador.

* We should be using `each` to confirm if an Object is `Enumerable`

I'll try to fix it Monday unless anyone else wants to try.

Original report: fog/fog-brightbox#17

tokengeek:
This is work to fix #138 

We check if the object passed to `Fog::Formatador` is `Enumerable` rather than checking for `map`

The class method is now used for all `Fog::Model#inspect` calls (whilst it used to just be used for collections).

This fixes fog/fog-brightbox#17

### DIFF
--- a/lib/fog/formatador.rb
+++ b/lib/fog/formatador.rb
@@ -15,8 +15,6 @@ module Fog
       string << "#{indentation}>"
     end
 
-    
-
     def self.display_line(data)
       ::Formatador.display_line(data)
     end
@@ -69,7 +67,7 @@ module Fog
     def self.nested_objects_string(object)
       nested = ""
       return nested if object.respond_to?(:empty) and object.empty?
-      return nested unless object.respond_to?(:map)
+      return nested unless object.is_a?(Enumerable)
       nested = "#{indentation}[\n"
       indent { nested << indentation + inspect_object(object) }
       nested << "#{indentation}\n#{indentation}]\n"
@@ -83,7 +81,7 @@ module Fog
     end
 
     def self.inspect_object(object)
-      return "" unless object.respond_to?(:map)
+      return "" unless object.is_a?(Enumerable)
       object.map { |o| indentation + o.inspect }.join(", \n#{indentation}")
     end
   end
--- a/spec/formatador_spec.rb
+++ b/spec/formatador_spec.rb
@@ -108,4 +108,47 @@ describe Fog::Formatador do
       Fog::Formatador.format(@subject).must_equal @expected
     end
   end
+
+  describe "when object responds to non-enumerable '#map'" do
+    before do
+      @member = Class.new(Fog::Model) do
+        def self.name
+          "IPAddress"
+        end
+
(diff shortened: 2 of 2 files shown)


geemus:
Thanks!

tokengeek:
Cheers @geemus

PR #141, opened 2015-03-24
Don't raise an exception if the final yield returns true

pdrakeweb:
Currently, if the final iteration in wait_for happens after timeout is exceeded and the yield returns true, an exception is still raised despite the successful block execution.  The included spec test fails with the existing code and passes with the new code, so it is a regression test, I guess.


plribeiro3000:
Thanks @pdrakeweb ! =)

geemus:
Good catch, thanks!

PR #149, opened 2015-04-09
Add Fog::Model equality based on identity

### LINKED ISSUE #148: `==` in Fog::Model currently compares actual objects; should it instead compare `identity`?
I'm working on implementing more rigorous tests for `fog-google`, and I was surprised to see that `==` isn't reimplemented in `Fog::Model` to compare `identity` or something similar.  In particular, I would expect that:

```
> instance = Fog::Compute[:provider].servers.create(params)
...
> Fog::Compute[:provider].servers.all.include? instance
```

would return `true`, but because `==` is still comparing literal Ruby objects, it returns false.

ihmccreery:
Closes #148

N.B. This implementation means that the following will happen:

```
> a = Fog::Model.new
> b = Fog::Model.new
> a == b
=> true
> a.save
> b.save
> a == b
=> false
```

It wasn't clear to me if two unsaved models should be equal or not, so I made a judgement call.  I'm definitely open to ideas.

### DIFF
--- a/lib/fog/core/model.rb
+++ b/lib/fog/core/model.rb
@@ -24,6 +24,14 @@ module Fog
       Fog::Formatador.format(self)
     end
 
+    def ==(o)
+      if (o.identity.nil? and self.identity.nil?)
+        o.object_id == self.object_id
+      else
+        o.class == self.class and o.identity == self.identity
+      end
+    end
+
     def reload
       requires :identity
 
--- a/spec/core/model_spec.rb
+++ b/spec/core/model_spec.rb
@@ -1,0 +1,29 @@
+require "spec_helper"
+require "securerandom"
+
+class FogTestModel < Fog::Model
+  identity  :id
+end
+
+describe Fog::Model do
+  describe "#==" do
+    it "is equal if it is the same object" do
+      a = b = FogTestModel.new
+      assert_equal a, b
+    end
+
+    it "is equal if it has the same non-nil identity" do
+      id = SecureRandom.hex
+      assert_equal FogTestModel.new(:id => id), FogTestModel.new(:id => id)
+    end
+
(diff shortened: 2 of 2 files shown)


geemus:
One edge case I can think of would be with nil identities. ie is it legit to say `compute.servers.new(...) == compute.servers.new(...)`? It seems alright in the case where identity exists, but in this case it seems weird, I think. What do you think?

ihmccreery:
Ooh, yeah, so even if there are other, *different*, attributes, as long as `identity` is `nil`, it still comes back true.  Bad news.  Okay, I'll change it so that if `identity` is `nil`, the objects must be the same object.

ihmccreery:
@geemus What are you thinking about this with the new changes?

geemus:
Hey, sorry, have just been a bit swamped this week. Looks good, thanks!

geemus:
@ihmccreery Hmm. Looks like this may be causing issues in some other tests, like https://travis-ci.org/fog/fog-aws/jobs/58885006#L324. This may just point to those tests being wrong, but it did highlight an oversight on our part. It is feasible that somebody would do `model == not_model`, and this doesn't handle that well (since it tries to call identity on the other object, which might not define that, as seen on travis. I've added a fallback to super equality checking in that case here: https://github.com/fog/fog-core/commit/d1e022ee3799d06d5a4cebf2a38865aaf884635d

I just did it quickly to try and fix the broken build, so it might be good to get more testing around this. Anyway, just wanted to make sure you were aware and let you add your thoughts. Thanks!

ihmccreery:
Gah, sorry about that.  Good call.  I've added another test that proves the functionality you're talking about: #152.

geemus:
No worries, I totally didn't consider it either, thanks!

PR #155, opened 2015-06-16
Fix account service to recognize all providers

mfbmina:
I was trying to use Fog::Account on fog-softlayer and I was always getting this message error:
``` softlayer has no account service (ArgumentError) ```

So, when I take a look at this service, I've seen that it always returning this message if provider was different from :stormondemand.

Then, I fixed it!

### DIFF
--- a/lib/fog/account.rb
+++ b/lib/fog/account.rb
@@ -11,6 +11,9 @@ module Fog
       if provider == :stormondemand
         require "fog/account/storm_on_demand"
         Fog::Account::StormOnDemand.new(attributes)
+      elsif providers.include?(provider)
+        require "fog/#{provider}/account"
+        return Fog::Account.const_get(Fog.providers[provider]).new(attributes)
       else
         raise ArgumentError, "#{provider} has no account service"
       end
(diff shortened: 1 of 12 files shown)


mfbmina:
If you can release a new version of fog-core it would be awesome! :smile:

plribeiro3000:
I believe you should do something in the lines of [this](https://github.com/PaeDae/nexus/pull/154) to make it compatible with the 2 supported implementations (`Fog::Account::Provider` and `Fog::Provider::Account`).

mfbmina:
@plribeiro3000 I did what you told me, and I updated all other models too. Hope it useful!

geemus:
Thanks!

geemus:
released v1.31.0

plribeiro3000:
Thanks @mfbmina !

mfbmina:
Thank you guys. I see that you made a fix commit on my PR. What was wrong?

geemus:
@mfbmina ah, there were some tests on fog/fog where it tried to lookup stuff like monitoring for services that don't actually define it (and expect to get a particular error back). Because of the way it was previously written it would just get `undefined method new for nil` instead of the nicer error saying that provider doesn't have that service. Hope that clarifies (basically it was never reaching the else, like it used to).

mfbmina:
Ok @geemus ! I asked just to learn more about fog! :smile:  thank you again!

geemus:
No problem, thanks for your help!

PR #157, opened 2015-07-14
Expose the name of the identity attribute

Ladas:
It's nice to have name of the identity attribute exposed as
public. So it can be used e.g. for selecting uniq elements
of the Fog::Collection


geemus:
Sounds good, thanks!

PR #160, opened 2015-09-20
add new optioin nested_aliases to attribute.

sawanoboly:
Hi, I added new option to attribute.

I would like to use value which in nested json as model attribute directly.

----

For example. 

When a provider returns nested key values like below.

```
{
      "Index": 0,
      "ID": "112700xxxxx",
      "Name": "example",
      "Description": "desc",
      "Settings": {
        "DNS": {
          "ResourceRecordSets": []
        }
      },
      "Status": {
        "Zone": "example.com",
        "NS": [
          "ns1.gslb2.sakura.ne.jp",
          "ns2.gslb2.sakura.ne.jp"
        ]
      }
}
```

We can use `:nested_aliases` to collect into attribute.

```
        attribute :zone, :nested_aliases => 'Status.Zone'
        attribute :nameservers, :nested_aliases => 'Stat.NS'
```

It creates below object.

```
=>   <Fog::DNS::SakuraCloud::Zone
    id="11270077xxxx",
    name="example.com",
    description="desc",
    zone="example.com",
    nameservers=["ns1.gslb2.sakura.ne.jp", "ns2.gslb2.sakura.ne.jp"],
```

### DIFF
--- a/lib/fog/core/attributes.rb
+++ b/lib/fog/core/attributes.rb
@@ -130,6 +130,7 @@ module Fog
         end
 
         self.class.nested_aliases.each_pair do |key, value|
+          next if attributes[value]
           attributes[value] = _get_nested_value(new_attributes, key.split("."))
         end
 
--- a/lib/fog/core/attributes.rb
+++ b/lib/fog/core/attributes.rb
@@ -197,8 +197,7 @@ module Fog
 
       # Dig value from new_attributes recursively
       # Returns String or Array
-      def _get_nested_value(attributes = {}, keys)
-        return nil unless attributes
+      def _get_nested_value(attributes = {}, keys = [])
         return attributes unless attributes.is_a?(Hash)
         if keys.length > 1
           _get_nested_value(attributes[keys.first], keys.drop(1))
--- a/lib/fog/core/attributes.rb
+++ b/lib/fog/core/attributes.rb
@@ -9,6 +9,10 @@ module Fog
         @aliases ||= {}
       end
 
+      def nested_aliases
+        @nested_aliases ||= {}
+      end
+
       def associations
         @associations ||= {}
       end
@@ -124,6 +128,11 @@ module Fog
             attributes[key] = value
           end
         end
+
+        self.class.nested_aliases.each_pair do |key, value|
+          attributes[value] = _get_nested_value(new_attributes, key.split("."))
+        end
+
         self
       end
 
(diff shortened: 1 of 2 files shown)


plribeiro3000:
I'm not sure about this feature. Maybe we just need a `Hash` type attribute?

sawanoboly:
Hmm... Hash type and currently default Type will not be the same?

I think that I can create symbolized hash from them. 
But, it will be kept strange nested structure which provided by the provider...

plribeiro3000:
Hmmm. You are right. The dafault will give you a hash object. So perhaps all you need are reader methods that will pick the data you want from the hash, like:

```ruby
module Fog
  module DNS
    class SakuraCloud
      class Zone < Fog::Model
        identity :id, :aliases => 'ID'
        attribute :name, :aliases => 'Name'
        attribute :description, :aliases => 'Description'
        attribute :settings, :aliases => 'Settings'
        attribute :status, :aliases => 'Status'

        def rr_sets
          settings.fetch('DNS', {}).fetch('ResourceRecordSets', [])
        end

        def zone
          status.fetch('Zone')
        end
      
        def nameservers
          status.fetch('NS')
        end
      end
    end
  end
end
```

Does the above work for you? Or you think we should support this in the DSL?

If you think we should i would like to think about another approach. This `nested` thing doesn't sound good to me, but HEY, thats just my opinion. =)

sawanoboly:
Ah, adding reader method sounds good :smiley: I'll try that approach.

sawanoboly:
BTW, Now, I was thinking another way. How about support Proc Object to collecting attribute?

plribeiro3000:
@sawanoboly Can you provide a sample of how the DSL would look like with a Proc?

sawanoboly:
For example, 

```
attribute :zone, :proc => Proc.new { self.status['Zone'] }
or 
attribute :zone, :proc => Proc.new { |res| res['Status']['Zone'] }
```

But, I can't decide that it is good way and works correctly.

plribeiro3000:
I don't know. It sounds a bit over complicated when all you want is to get a key from a hash attribute.
Maybe we can work on a `Hash` attribute which gives you a better result then the `Default` one?

sawanoboly:
The reason why I worry, I have the following.

- `zone` must need to create new zone.
- `rr_sets` is Array of Hash. And it must keep all record to update or reduce.

plribeiro3000:
So you need to validate the an attribute that is inside the `Hash`?
This is a hard problem to solve and now i got what you want but i still feel that we should go with the `Hash` attribute and work on top of it. What do you think?

sawanoboly:
I agree to close this PR. And I'll implement getter and setter to my model until I have come up with new idea.
Thanks,

geemus:
It seems like a hard case, but I think I would agree that getter/setter are probably the right place to start. Should get what you need without introducing too much complexity that we are not sure about yet.

sawanoboly:
Thanks @plribeiro3000 @geemus 

I've released  https://github.com/fog/fog-sakuracloud/pull/15 . It supports SakuraCloud DNS.

plribeiro3000:
sure @sawanoboly . I'm glad it worked!

PR #161, opened 2015-09-25
Locking the net-ssh for Ruby <2.0

Temikus:
Looking at the dep graph, it's fog-core's dependency net-ssh, which just updated a couple of hours ago:
https://rubygems.org/gems/net-ssh/
3.0.1 - September 25, 2015 (177 KB)
This breaks gem deps on Ruby <2.0

### DIFF
--- a/fog-core.gemspec
+++ b/fog-core.gemspec
@@ -23,7 +23,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency("formatador", "~> 0.2")
   spec.add_dependency("mime-types")
   spec.add_dependency("net-scp", "~> 1.1")
-  spec.add_dependency("net-ssh", ">= 2.1.3")
+  spec.add_dependency("net-ssh", ">= 2.1.3", "< 3.0")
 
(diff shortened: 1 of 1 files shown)


plribeiro3000:
Thank you!

tokengeek:
I think to avoid the problems we had with `nokogiri` (dropping 1.8.7 in one version but the next version being required by Rails 4) - we should replace this change with documentation and a 1.9.3 Gemfile.

Basically, if you are running an app on Ruby 1.9.3, you probably need to be stricter with your own dependencies rather than us limiting new applications again.

Our CI server is filled with errors due to the original change and will be until a new release is made because all the available gems still have the optimistic dependency.

I'll create a PR for those changes but mentioning it here.

tokengeek:
PR as mentioned above https://github.com/fog/fog-core/pull/163

plribeiro3000:
agree with you here @tokengeek . It should be way easier to the final users to be strict then us maintaining all this versions support. =s

PR #163, opened 2015-09-28
Cleanup dependencies

tokengeek:
Tweaking the Gemfiles and travis settings in order to make
the dependencies as open as possible.

Limiting `nokogiri` for older rubies caused problems for users with
Rails 4.0 applications so we should prioritise supported versions
of Ruby and document workarounds.


tokengeek:
Hoping to add some documentation before merging, then removing
the dependency added to the `gemspec`

Unfortunately I can see we either update all the Gemfiles and travis.yml
which is the right thing in my opinion, or add a restriction for newer Rubies
which is a simpler (short term fix).

.travis.yml
```diff
@@ -19,26 +11,21 @@ script: bundle exec rake travis
...
-      gemfile: Gemfile
-    - rvm: 1.9.2
-      gemfile: Gemfile.1.8.7
-    - rvm: 1.9.3
       gemfile: Gemfile.1.8.7
-    - rvm: 2.0.0
+    - rvm: jruby-18mode
       gemfile: Gemfile.1.8.7
+    - rvm: 1.9.2
+      gemfile: Gemfile.1.9.3
```
plribeiro3000: I believe we just dropped support to `1.9.2` at fog `2.0.0`. Could we remove it here?

tokengeek: Yep. We probably should. I'll update.

icco:
@tokengeek + @plribeiro3000 I merged master into this and updated travis to be closer to Fog. Thoughts?

plribeiro3000:
LGTM!

plribeiro3000:
Travis seems to be having problems building the ruby binaries. =s

geemus:
Bringing this in. Looks like there is a broken test, but that it was pre-existing condition.

PR #166, opened 2015-11-09
loosen requirements around net-ssh/net-scp and ruby version

geemus:
first stab at making 1.9 still a thing. Might even remove the ruby require to allow this to more easily go onto 1.x series (and then get back onto the 2.x train).

### DIFF
--- a/fog-core.gemspec
+++ b/fog-core.gemspec
@@ -17,14 +17,12 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^bin/}) { |f| File.basename(f) }
   spec.test_files    = spec.files.grep(%r{^(test|spec|features)/})
   spec.require_paths = ["lib"]
-  spec.required_ruby_version = '>= 2.0'
+  spec.required_ruby_version = '>= 1.9'
 
   spec.add_dependency("builder")
   spec.add_dependency("excon", "~> 0.45")
   spec.add_dependency("formatador", "~> 0.2")
   spec.add_dependency("mime-types")
-  spec.add_dependency("net-scp", "~> 1.1")
-  spec.add_dependency("net-ssh", ">= 2.1.3", "< 3.0")
 
   spec.add_development_dependency("coveralls")
   spec.add_development_dependency("minitest")
--- a/lib/fog/compute/models/server.rb
+++ b/lib/fog/compute/models/server.rb
@@ -65,7 +65,6 @@ module Fog
       end
 
       def scp(local_path, remote_path, upload_options = {})
-        require "net/scp"
         requires :ssh_ip_address, :username
 
         Fog::SCP.new(ssh_ip_address, username, ssh_options).upload(local_path, remote_path, upload_options)
@@ -74,14 +73,12 @@ module Fog
       alias_method :scp_upload, :scp
 
       def scp_download(remote_path, local_path, download_options = {})
-        require "net/scp"
         requires :ssh_ip_address, :username
 
         Fog::SCP.new(ssh_ip_address, username, ssh_options).download(remote_path, local_path, download_options)
       end
 
       def ssh(commands, options = {}, &blk)
-        require "net/ssh"
         requires :ssh_ip_address, :username
 
         options = ssh_options.merge(options)
--- a/lib/fog/core/scp.rb
+++ b/lib/fog/core/scp.rb
@@ -40,7 +40,12 @@ module Fog
 
     class Real
       def initialize(address, username, options)
-        require "net/scp"
+        begin
+          require "net/scp"
+        rescue LoadError
+          Fog::Logger.warning("'net/ssh' missing, please install and try again.")
+          exit(1)
+        end
 
         key_manager = Net::SSH::Authentication::KeyManager.new(nil, options)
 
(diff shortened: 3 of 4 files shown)


plribeiro3000:
:+1:

geemus:
@plribeiro3000 thanks for taking a look. We'll probably have to sweep through a bunch of other stuff to look for more requires, but hopefully it is mostly (if not wholly) limited to this core part. Will be nice to put all this behind us!

plribeiro3000:
Sure it will!

Temikus:
@geemus It looks like this causes issues due to the rescue block in ServicesMixin.
Relevant bit in `/lib/fog/core/services_mixin.rb`
```
    def new(attributes)
      attributes    = attributes.dup # Prevent delete from having side effects
      provider      = attributes.delete(:provider).to_s.downcase.to_sym
      provider_name = Fog.providers[provider]

      raise ArgumentError, "#{provider} is not a recognized provider" unless providers.include?(provider)

      require_service_provider_library(service_name.downcase, provider)
      spc = service_provider_constant(service_name, provider_name)
      spc.new(attributes)
    rescue LoadError, NameError  # Only rescue errors in finding the libraries, allow connection errors through to the caller
      raise ArgumentError, "#{provider} has no #{service_name.downcase} service"
    end
```

Repro:
1) Check out fog-google.
2) Modify Gemfile to:
```
source "https://rubygems.org"

gem 'fog-core', :git => 'git@github.com:fog/fog-core.git', :branch => 'loosen-requires'

# Specify your gem's dependencies in fog-google.gemspec
gemspec
```
3) Run this example:
```
require 'bundler'
require 'readline'

Bundler.require(:default, :development)
WebMock.disable!

connection = Fog::Compute.new({ :provider => :google })
puts "Connected!" if connection
pp connection
```

EXPECTED RESULT:
```
λ bundle exec ruby example_connect.rb
Connected!
```
ACTUAL RESULT:
```
λ bundle exec ruby example_connect.rb
/Users/temikus/.rbenv/versions/2.1.6/lib/ruby/gems/2.1.0/bundler/gems/fog-core-7a87027df766/lib/fog/core/services_mixin.rb:18:in `rescue in new': google has no compute service (ArgumentError)
	from /Users/temikus/.rbenv/versions/2.1.6/lib/ruby/gems/2.1.0/bundler/gems/fog-core-7a87027df766/lib/fog/core/services_mixin.rb:8:in `new'
	from /Users/temikus/.rbenv/versions/2.1.6/lib/ruby/gems/2.1.0/bundler/gems/fog-core-7a87027df766/lib/fog/compute.rb:62:in `new'
	from example_connect.rb:7:in `<main>'
```

As soon as we add net-ssh back:
```
source "https://rubygems.org"

gem 'fog-core', :git => 'git@github.com:fog/fog-core.git', :branch => 'loosen-requires'
gem 'net-ssh'

# Specify your gem's dependencies in fog-google.gemspec
gemspec
```
, everything is peachy:
```
λ bundle && bundle exec ruby example_connect.rb
Connected!
```

geemus:
@Temikus good call. I suspect it might be a direct require of `net-ssh` or `net-scp` somewhere inside the google provider though? I haven't started that survey yet to fix in the various providers, but I think that is more likely to be at fault given my current understanding of things anyway. Does that help/make sense? I hope to go hunting for such references soon, but I don't think it would result in this (the branch does `exit(1)`, so it should error, but not in this way).

Temikus:
@geemus This is true that this concrete example fails due to `require "net/ssh/proxy/command"` in `lib/fog/google/models/compute/server.rb`.

However, I feel this may make our user experience somewhat confusing. People starting out may not have any idea as to why they receive the message of `google has no compute service`. I haven't dug that deep into fog-core, so I'm not sure how easy it is to fix or work around, but I think it would be better to fix the message in core first before going hunting for all references in the providers. Makes sense?

I may be missing something though - let me know if so.

geemus:
Yeah. I see these changes as necessary but not sufficient. I think many providers will also need to be updated with similar optional requires (and probably released first, then when this release happens it shouldn't break anything I think/hope). I just haven't managed to start working on the various providers just yet...

Temikus:
Understood, thanks! I can take a look at fog-google when I have some spare cycles if you need help.
Though I need to note again that we already did quite a bit of changes in there that _may_ be incompatible with 1.9.3, so:
- I can make the load routine optional.
- I'm not sure whether that will actually allow us to run on 1.9.3 anymore.
Should I check it out?

plribeiro3000:
@Temikus I believe you could just lock the version at `fog` and call it a day.

geemus:
Bringing this in. Will follow up on the other bits ASAP.

geemus:
I locked in fog, at least for the time being.

nurse:
We finished the support for Ruby 1.9.3 on February.
https://www.ruby-lang.org/en/news/2015/02/23/support-for-ruby-1-9-3-has-ended/

plribeiro3000:
@nurse Yeah, thats why we dropped support for it but we had to get it back due to the amount of other libs that use `fog` and that still support `1.9.3`. One example are the LTS distros like `ubuntu` and `redhat` that still have `1.9.3` as the stable ruby version.

PR #172, opened 2015-12-09
Make digitalocean v2 api default

h0lyalg0rithm:
API v1 is at eol.More information here
https://developers.digitalocean.com/documentation/v1/ 

closes fog/fog#3782

### DIFF
--- a/lib/fog/compute.rb
+++ b/lib/fog/compute.rb
@@ -42,12 +42,12 @@ module Fog
       when :digitalocean
         version = attributes.delete(:version)
         version = version.to_s.downcase.to_sym unless version.nil?
-        if version == :v2
+        if version == :v1
+          error_message = "DigitalOcean V1 is deprecated.Please use `:version => :v2` attribute to use Next Gen Cloud Servers."
+          raise error_message
+        else
           require "fog/digitalocean/compute_v2"
           Fog::Compute::DigitalOceanV2.new(attributes)
-        else
-          require "fog/digitalocean/compute"
-          Fog::Compute::DigitalOcean.new(attributes)
         end
       when :stormondemand
         require "fog/compute/storm_on_demand"


plribeiro3000:
@h0lyalg0rithm What if you just check for `v1` instead and anything else falls in `v2`?
This would count '' as `v2` as well.

petems:
@plribeiro3000 Would an idea to add deprecation warning or even a full error when someone is using `v1`, as it literally won't work anymore, would be good to explain that. Otherwise someone using some old code they found will get a bunch of unexplained errors.

plribeiro3000:
@petems I didn't quite get what you mean.

But my point is that we should have warnings unless the user is using `v2`. Otherwise it will just stop working as soon as `v1` goes down. Anything but `v2` should have warning logs.

petems:
@plribeiro3000 Sorry I explained that badly, looking at the code it already does what I asked :sweat: 

FYI  v1 is already down, was November 9th IIRC: 

```
API v1 has reached end-of-life. All subsequent requests to API v1 will respond with HTTP 410 (Gone) and an error message:
```

h0lyalg0rithm:
@petems I just changed the code to check if the version is 2.
If the user doesnt specify the version number it will use version 2 by default.

plribeiro3000:
@h0lyalg0rithm better i think. Just require the old version if it is `v1`.
I know it wont work but at least it needs to throw a error different of undefined method.

geemus:
Yeah, in addition to a warning, we probably should go ahead and error also to be more explicit.

h0lyalg0rithm:
Thanks for the feedback. I have changed the code to raise an error if they use version v1 for the digitalocean api.

geemus:
Looks good, thanks!

h0lyalg0rithm:
Thanks for accepting the PR

plribeiro3000:
@h0lyalg0rithm Sorry if i missed your PR. We are lucky we have @geemus . =)

Thanks for your efforts!

PR #174, opened 2016-01-10
Don't eagerload auto-loaded service classes in Fog::Mock.reset.

nirvdrum:
If the service classes haven't been loaded yet, they cannot have been used in any mocked tests, so there's nothing to be gained by loading them to clear out state. Eager-loading isn't just an efficiency issue -- eagerly loading classes may unnecessarily conflict with the rest of the dependency graph and can cause contrived breakages if a service doesn't run particularly well on a given platform.

### DIFF
--- a/lib/fog/core/mock.rb
+++ b/lib/fog/core/mock.rb
@@ -91,10 +91,13 @@ module Fog
     def self.reset
       mocked_services = []
       Fog.constants.map do |x|
+        next if Fog.autoload?(x)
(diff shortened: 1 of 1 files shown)


nirvdrum:
While I can merge this on my own, I'd appreciate another set of eyes. I think my analysis is correct and it appears to fix the issues I've been seeing in Travis.

geemus:
Sounds good, thanks!

PR #175, opened 2016-02-25
Remove hp from the providers list

h0lyalg0rithm:
Remove hp from the list since their service has shutdown and fog/fog no longer has it.
Refer fog/fog#3854


plribeiro3000:
Thanks!

geemus:
Thanks!

PR #176, opened 2016-03-04
Re-raise LoadError rather than exiting

glittershark:
Re-raise LoadError when unable to require `mime/types` rather than exiting - the latter causes any application *using* fog to exit, which is a really crappy thing to run in to on accident for a web app


geemus:
Thanks!

PR #179, opened 2016-03-30
Adds Introspection module

gildub:
Openstack offers introspection with coming introspection service
(fog-openstack). The latter needs the mixin for Mock to work.


geemus:
Thanks!

PR #181, opened 2016-04-01
change service not found error type to Fog::Errors::Notfound

zhitongLBN:
Use `Notfound` error instead of `ArgumentError` when the service was not found,
It could be very nice and more comprehensive.

Project like MIQ try to catch NotFound for not found service in provider catalog
https://github.com/Ladas/manageiq/blob/c836fdf2567122193ec54d237bc07df814abe407/gems/pending/openstack/openstack_handle/handle.rb#L75

### DIFF
--- a/lib/fog/core/services_mixin.rb
+++ b/lib/fog/core/services_mixin.rb
@@ -15,7 +15,7 @@ module Fog
       spc = service_provider_constant(service_name, provider_name)
       spc.new(attributes)
     rescue LoadError, NameError  # Only rescue errors in finding the libraries, allow connection errors through to the caller
-      raise ArgumentError, "#{provider} has no #{service_name.downcase} service"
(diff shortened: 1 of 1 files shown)


geemus:
Sounds good, thanks!

zhitongLBN:
@geemus Finally @Ladas from MIQ ask for type of error more specify like `ServiceNotFound`,
 If you are agree with this, I will remake a new PR, sorry for confusing

Ladas:
@zhitongLBN I think you can just sent another PR changing it to something more specific? No need to revert afaik

geemus:
@zhitongLBN No need to revert I don't think. This was an improvement, we could just go even further. If you could just make a new PR that has `ServiceNotFound` which probably is a child of NotFound to be thorough we can use that. Make sense?

PR #182, opened 2016-04-04
Revert "change service not found error type to Fog::Errors::Notfound"

zhitongLBN:
Reverts fog/fog-core#181

@geemus Finally @Ladas from MIQ ask for type of error more specify like `Fog::Error::ServiceNotFound`,
If you are agree with this, I will remake a new PR, sorry for confusing


geemus:
I don't think we need to revert. This is better than before, we can maybe just do even better than this.

PR #185, opened 2016-05-06
Adds NFV module

Ladas:
OpenStack have a new NFV service called Tacker


plribeiro3000:
@Ladas Is there another provider that can use this as well? I believe we have a standard to add new Services only when 2 or more providers need it.

cc/ @geemus

Ladas:
@plribeiro3000 Network Function Virtualization(NFV) is a term heard everywhere, so I would say so. :-) But for now, just OpensStack, I haven't investigated other providers TBH.

I've seen Introspection being added for OpenStack, so I guess OpenStack counts for 2? :-)

seanhandley:
I don't know of any public cloud services offering NFV so maybe OpenStack's just ahead of the curve

plribeiro3000:
@Ladas Thats the main point on that i guess. To make sure there is at least 3 providers that need it.
But we can always make a different call based on the circunstances.

@geemus Might have something to add.

Ladas:
@geemus do you think we could merge and release? NFV is the new cool. :-)

geemus:
The push behind waiting would be that if we move forward now there is a danger that we would bind the interface too explicitly to openstack in particular, vs having it abstract across differing implementations. We should be able to support/release NFV without it being at the top level I would think? Any reason not to just do that?

Ladas:
Actually, we always had problems when the namespace were not top level. Without defining the right module, I can't use Fog::NFV[:openstack], or is there some hack for this?

Btw. the code will be just in fog-openstack, so there should not be any limitations with interfaces? Just that if anybody has a service that allows to work with NFV devices, it should go in Fog::NFV.

plribeiro3000:
@Ladas I guess the point is that we don't know that yet so any work done now is a prediction on what it would become. In the future `NFV` might have another name that all providers are using and then we will have kind of confusion because providers work in a different way.

Can't you have this inside `Compute` for now and in a breaking version move it to `NFV`?
This would be one of the benefits of having each provider in a separate gem.

geemus:
Yeah, Openstack::NFV could do whatever it wants (there are many services more like this on different providers), but the top level stuff is more like compute/storage/etc where we have tried to have the same basic/core interface across all providers. @Ladas does that help/make more sense?

Ladas:
@plribeiro3000 not sure if putting this under another service will work, since it has it's own endpoint. Or can we do multiple endpoints per service? I haven't seen that used.

@geemus I guess, but I haven't really found a way how to make Fog::NFV[:openstack] work, without putting the module into fog-core :-)

gildub:
Openstack is not alone, it seems there are at least 2 other providers offering NFV features:
- Google Cloud Platform: Andromeda
- VMware: vCloud NFV Platform
That makes it a good candidate for fog-core.

I vote for NFV name:
NFV seems to be a good name because it's already an abstract of network architectures, under the umbrella of the Network Virtualization concept. We cannot use Network Virtualization as a name, it's too broad. Well, it seems the `function` part of NFV is really important, that's the programming part to put together functions to assemble network services.

Ladas:
@plribeiro3000 there you go, count == 3, thanks @gildub 

If possible, I would prefer to have fog-core NFV, it's a well established term in SDN world now. :-)

geemus:
Fair enough, it was news to me I guess. Just pays to be cautious vs having to deal with a lot of stuff that might be hard to change in the future. Let's go ahead and roll forward with this then. Thanks.

Ladas:
@geemus \o/ 

Could you also do a fog-core release, please? I just finished the fog-openstack code, so I will need that to pass the CI. Thank you. :-)

geemus:
Done, v1.39.0

Ladas:
@geemus awesome, thank you

Ladas:
did it fail maybe? I don't see 1.39 in https://rubygems.org/gems/fog-core/versions/1.38.0

geemus:
Weird, I guess it must have. I tried again and it appears to have gone through this time.

Ladas:
@geemus cool, I see it now, thank you :-)

plribeiro3000:
@Ladas @gildub Did not know that also. =)

Agree with @geemus words on being cautious though. 👍

PR #187, opened 2016-05-16
Minitest match_schema helpers assertion/expectation

gildub:
Adds
- assert_match_schema assertion
- must_match_schema expectation

which is based upon Fog::Schema::DataValidator

This is the equivalent of data_matches_schema (shindo)


lib/fog/test_helpers/minitest/assertions.rb
```diff
@@ -0,0 +1,35 @@
+require "fog/schema/data_validator"
+
+# Format related hackery
+# allows both true.is_a?(Fog::Boolean) and false.is_a?(Fog::Boolean)
+# allows both nil.is_a?(Fog::Nullable::String) and ''.is_a?(Fog::Nullable::String)
+module Fog
```
Ladas: do you think you could extract L6-L25 to another file and require it here? The same for https://github.com/fog/fog-core/blob/master/lib/fog/test_helpers/formats_helper.rb Otherwise it's a lot of duplicated code :-)

Something like TypesHelper, or whatever naming you like. Even the FormatsHelper is good, but without the shindo code.

geemus: Yeah, consolidating vs duplicating sounds like a good plan to help reduce the likelihood of drift in the future. Thanks!

Ladas:
from me, just one comment, otherwise looks good :+1:

lib/fog/test_helpers/minitest/assertions.rb
gildub: Yeah I've been thinking about that to. I thought the code in Shindo to go ultimately go away. Meanwhile that's going to take time so better do it. I'm creating a file for the common code.

gildub: Yeah I've been thinking about that too. I thought the code in Shindo to ultimately go away. Meanwhile that's going to take time so better do it. I'm creating a file for the common code.

geemus:
Very true. It hasn't changed much either, so maybe that is premature. Let's just bring this in and keep an eye on it. Thanks!

PR #192, opened 2016-05-28
Fixing namespace constant detection

Temikus:
- Disabled inherit option on `const_get` in `service_provider_constant` so it doesn't search the ancestors and `Object`.

This should fix https://github.com/fog/fog-google/issues/123

/CC  @geemus @tokengeek @plribeiro3000 PTAL, since I'm less familiar with `fog-core` code.


geemus:
LGTM, thanks!

plribeiro3000:
:+1:

Great finding.

PR #200, opened 2016-10-31
Moves Mock class to fog-mock gem

gildub:
Can be merged only once fog-mock first version is available.


fog-core.gemspec
```diff
@@ -13,7 +13,7 @@ Gem::Specification.new do |spec|
   spec.homepage      = "https://github.com/fog/fog-core"
   spec.license       = "MIT"
 
-  spec.files         = `git ls-files`.split($INPUT_RECORD_SEPARATOR)
+  spec.files         = `git ls-files -z`.split("\x0")
```
plribeiro3000: Why this change?

lib/fog/core.rb
```diff
@@ -40,7 +40,6 @@
 require File.expand_path('../core/hmac', __FILE__)
 require File.expand_path('../core/logger', __FILE__)
 require File.expand_path('../core/model', __FILE__)
-require File.expand_path('../core/mock', __FILE__)
```
plribeiro3000: I believe that rather the removing you need to change the path to the new one like `fog/mock` no?

lib/fog/core/errors.rb
```diff
@@ -11,8 +11,6 @@ def self.slurp(error, message = nil)
       end
     end
 
-    class MockNotImplemented < Fog::Errors::Error; end
```
plribeiro3000: I'm still not sure if this exception should be extracted as well. Even that is being used only by Mock it is still an exception and split it might make it harder to figure out stuff later. What do you think?

lib/fog/core/scp.rb
```diff
@@ -1,43 +1,5 @@
 module Fog
   module SCP
-    def self.new(address, username, options = {})
-      if Fog.mocking?
-        Fog::SCP::Mock.new(address, username, options)
-      else
-        Fog::SCP::Real.new(address, username, options)
-      end
-    end
```
plribeiro3000: I believe we still need this.

lib/fog/core/service.rb
```diff
@@ -295,22 +273,6 @@ def camel_case_collection_name(collection)
...
-                Fog::Mock.not_implemented
-              end
-            EOS
-          end
-        end
-      end
-
       # Requires the correct file for an item (collection, model, or request).
       #
       # @param [Array] item
```
plribeiro3000: I'm not sure we should touch this logic yet. We need to be fully backwards compatible for now meaning we can do this breaking change yet.
As the first release we need to support the same features and behave the same way. Thats the best approach to make sure an extraction wont break anything.
Can you undo those changes please?

spec/spec_helper.rb
```diff
@@ -13,6 +13,7 @@
 
 $LOAD_PATH.unshift "lib"
 require "fog/core"
+require "fog/mock"
```
plribeiro3000: This sounds amazing in the final user tests. =)

Just a plain `require 'fog-mock'` at the beginning of the test helper. 👍

fog-core.gemspec
gildub: Good question, why are gem specs depending on some external environment value? Could not get the test running.

lib/fog/core.rb
gildub: No because this is an external gem, it requires a gem not a file. The gem is added to gem specs dependencies.

lib/fog/core/errors.rb
gildub: Yes I was thinking the same thing too. As you said it's used only by Mock, therefore it should be in its scope.

fog-core.gemspec
plribeiro3000: Dunno. This code predates my work on fog.

lib/fog/core/scp.rb
gildub: I think we do need that on the fog-mock side to be able to create Mock or Real classes when testing but not on the fog-core side where only Real objects should be living.

lib/fog/core/errors.rb
plribeiro3000: But it is still an exception. I believe this extraction is unnecessary and make it harder to maintain the codebase. Can we leave this code here?

lib/fog/core/service.rb
plribeiro3000: All the changes in this file. =)
We need it. We can't just rip it off like that. We need to be backwards compatible.

lib/fog/core/scp.rb
plribeiro3000: Thats true when we finish the decoupling. As for now we need to keep it otherwise we will break the library for a ton of users.

plribeiro3000:
@gildub Lets hold onto this a little bit. @geemus did comment about this agreeing that we might need to push one mock gem per extracted provider and thats not intended.
Let wait for his answer on my last comment.

lib/fog/core/errors.rb
gildub: Sure, consider it restored.

lib/fog/core/service.rb
gildub: Agreed with backward compatibility approach. As I said above, the difference now is that when a service is required, the Real class is called not any more Mock.

gildub:
@plribeiro3000, 

Sure, let's wait.

I think each sub project has the choice to extract Mock or not. Meanwhile if they do it will have to end up in a fog-mock-<provider> repo unless they choose to put it into the test structure.

fog-core.gemspec
tokengeek: FYI - `$/` was part of the orignal Bundler generated template.

`$INPUT_RECORD_SEPARATOR` is defined in the standard `English` library and was replaced by Rubocop to avoid having Perlism like `$/`

Presently `bundle gem` generates this line as:

```
spec.files = `git ls-files -z`.split("\x0").reject { |f| f.match(%r{^(test|spec|features)/}) }
```

You can either `require "English"` for these readable version to be defined or replace it with the latest Bundler line.

plribeiro3000:
This can be close now that we abandoned the `fog-mock` idea.

Comment back if you want do discuss it further.

Thanks for everything!

Issue #198, opened 2016-10-28
The mocking bird


gildub:
Can we please remove mock?

There are several reasons to remove the Mock class or at least move it away into unit tests.
Also, I believe I'm not the only one thinking in such way about the mock, so it seems to be the time to discus it.

Looking back, some time ago, I was thinking having unit testing capability close to the source was a good idea, meanwhile experience proved otherwise.

The mock class is more a burden than of any help.
The main issue, as far as I can see, is because the Service class is in essence only a conduit to cloud APIs. So there no real need to extensively mock them, by doing so, it makes it wrong and is  waste of time to write tests that are 'unrealistic' tests. To make that crystal clear, I've used my own service replacing the default ones, and guess what, not matter what I do, the tests are always GREEN. 

Fundamentally, once the service pattern is tested for once and for good, the API service which consist of sending a path and a hash(.to_json) and getting a response, leaves the rest to integration tests like VCR. The latter part actually tests an API works.

Also the unit test code shouldn't appear in production. Normally when packaging the gem, the test folder(s) is left aside. Well that's impossible to do because the Mock class is embedded into the Service subclasses.

So, can we, please?

Temikus:
I think this is relevant to #12

gildub:
@Temikus, yes removing any tests using Mock would have to be done too.

plribeiro3000:
@gildub There is interesting points here for sure.
But there is also the other side of the coin. Mock might be a pain for use to maintain in the source code but when it works (not all apis are available in mock mode) it helps who are developing using `fog` by a lot. I know that because i've done it before and it really helped.
Another point here is that the code you were using might not be in a good shape. I've got my code right so many times because of mock, either because i was not using the right id (so the object did not exist) or just because i was using the wrong signature.
There is also a lot of people out there that might be using this feature so we can't just drop it considering it is a great deal to `fog` and in my opinion the major advantage over other libraries where i need to craft live tests or don't have tests at all.

So i believe that if done right, the Mock feature helps a lot. What we might need is to make sure they work as they should and maybe extract Mock into `fog-mock` so we can load it in development and test environments only.

Also i believe that your real problem isn't the feature (Mock) but some code that is not up to date or well crafted, so removing the feature completely is not the right answer IMHO.

So i suggest we look at this from other angle. Which feature, code, service, provider did not work for you?

gildub:
@plribeiro3000, 

[...]

PR #201, opened 2016-11-01
Moves Mock to test zones

gildub:
The Mock structure, Mock module and the Service class logic handling
Mocks are now into lib/fog/test_helpers where they can be reused by
other fog-<providers>.

SCP and SSH mock classes have also been moved to spec/ as they aren't
part of the Mock structure itself but just consuming it.

https://github.com/fog/fog-core/issues/198


gildub:
@plribeiro3000, seems like there are deps issue. Json >= 2 is not compatible with Ruby < 2.
Are you facing the same issues?

gildub:
This one will more likely be applied to a coming Version 2 branch.

plribeiro3000:
@gildub Sorry for the late answer. Busy days.

Thanks for catching up.

PR #204, opened 2017-04-02
Add file yml caching to models

shaiguitar:
Hey @geemus regarding the caching issue @ https://github.com/fog/fog-aws/issues/354#issuecomment-286789702

To take care of before merging:

- [x] How to safely assume a service identifier? Currently using region just as a placeholder.
- [x] Add downcase for case insensitive platforms.
- [x] Documentation on public methods.
- [x] Ruby not use /tmp tmpdir change.
- [x] Other feedback.

Works with a test CLI application using this fork. You can see usage in the test file, the gist is:

- Expiring cache:
```
[1] pry> Fog::Cache.expire_cache!(Fog::Compute::AWS::Server, compute)
=> ["/tmp/Fog_Compute_AWS_Real/us-east-1/Fog_Compute_AWS_Server"]
```

- Raises on empty cache:
```
[2] pry> ss = Fog::Cache.load(Fog::Compute::AWS::Server, compute)
Fog::Cache::CacheNotFound: Fog::Cache::CacheNotFound
from /Users/shai/r/fog-core/lib/fog/core/cache.rb:15:in `load'
```

- Dumping an instance:
```
[1] pry> s = c.servers.sample; s.id
=> "i-0569a70ae6f47d229"
[2] pry> s.cache.dump
=> 2371
```

- On disk:
```
shai@adsk-lappy ~   % tree /tmp/Fog_Compute_AWS_Real/us-east-1/Fog_Compute_AWS_Server/
/tmp/Fog_Compute_AWS_Real/us-east-1/Fog_Compute_AWS_Server/
└── i-0569a70ae6f47d229.yml
```

- Loading cache in different session:
```
# Not sure if there's a way to determine model class from collection class then
# this call would read nicer (returns collection since we don't have identifiers, just globbing).

[1] pry> instances = Fog::Cache.load(Fog::Compute::AWS::Server, compute)
...
[2] pry> instances.first.id
=> "i-0569a70ae6f47d229"
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,4 +1,3 @@
 source 'https://rubygems.org'
-
 gemspec
 
--- a/fog-core.gemspec
+++ b/fog-core.gemspec
@@ -21,7 +21,9 @@ Gem::Specification.new do |spec|
   spec.add_dependency("builder")
   spec.add_dependency("excon", "~> 0.49")
   spec.add_dependency("formatador", "~> 0.2")
+  spec.add_dependency("xmlrpc") if RUBY_VERSION.to_s >= "2.4"
 
+  spec.add_development_dependency("tins") if RUBY_VERSION.to_s > "2.0"
   spec.add_development_dependency("coveralls")
   spec.add_development_dependency("minitest")
   spec.add_development_dependency("minitest-stub-const")
--- a/lib/fog/core/cache.rb
+++ b/lib/fog/core/cache.rb
@@ -153,9 +153,11 @@ module Fog
     end
 
     def self.sandbox
+      raise CacheDir.new("Must set an explicit sandbox for this cache. Example: 'serviceX-regionY'") unless sandbox_name
+
       @sandbox = if @sandbox_name
                    safe_path(File.expand_path("#{SANDBOX_PREFIX}/#{@sandbox_name}"))
-                 end || (raise CacheDir.new("Must set an explicit sandbox for this cache. Example: 'serviceX-regionY'"))
+                 end
     end
 
     def self.sandbox_name=(name)
@@ -166,11 +168,10 @@ module Fog
       @sandbox_name
     end
 
-
     # The path/namespace where the cache is stored for a specific +model_klass+ and +@service+.
     def self.namespace(model_klass, service)
-      name = File.join(self.sandbox, service.class.to_s, model_klass.to_s)
-      name = safe_path(name)
+      ns = File.join(self.sandbox, service.class.to_s, model_klass.to_s)
+      ns = safe_path(ns)
     end
 
     def self.safe_path(klass)
--- a/lib/fog/core/cache.rb
+++ b/lib/fog/core/cache.rb
@@ -131,6 +131,9 @@ module Fog
         Fog::Logger.warning("Found duplicate items in the cache. Expire all & refresh cache soon.")
       end
 
+      # Fog models created, free memory of cached data used for creation.
+      @memoized = nil
+
       uniq_loaded
     end
 
@@ -145,11 +148,12 @@ module Fog
       FileUtils.rm_rf(namespace(model_klass, service))
     end
 
-    # loads yml cache from path on disk
+    # loads yml cache from path on disk, used
+    # to initialize Fog models.
     def self.load_cache(path)
-      @cache ||= {}
-      return @cache[path] if @cache[path]
-      @cache[path] = YAML.load(File.read(path))
+      @memoized ||= {}
+      return @memoized[path] if @memoized[path]
+      @memoized[path] = YAML.load(File.read(path))
     end
 
     def self.namespace_prefix=(name)
--- a/lib/fog/core/cache.rb
+++ b/lib/fog/core/cache.rb
@@ -1,0 +1,210 @@
+# coding: utf-8
+require "fileutils"
+require "yaml"
+require "tmpdir"
+
+module Fog
+  # A generic cache mechanism for fog resources. This can be for a server, security group, etc.
+  #
+  # Currently this is a on-disk cache using yml files per-model instance, however
+  # there is nothing in the way of extending this to use various other cache
+  # backends.
+  #
+  # == Basic functionality
+  #
+  # set the namespace where this cache will be stored:
+  #
+  # Fog::Cache.sandbox_name = "service-account-foo-region-bar"
+  #
+  # cache to disk:
+  #
+  #   # after dumping, there will be a yml file on disk:
+  #   resouce.cache.dump
+  #
+  #   # you can load cached data in from a different session
+  #   Fog::Cache.load(Fog::Compute::AWS::Server, compute)
+  #
+  #   # you can also expire cache (removes cached data assocaited with the resources of this model associated to the service passed in).
+  #   Fog::Cache.expire_cache!(Fog::Compute::AWS::Server, compute)
+  #
+  # == More detailed flow/usage
+  #
+  # Normally, you would have a bunch of resources you want to cache/reload from disk.
+  # Every fog model has a +cache+ object injected to accomplish this. So in order to cache a server for exmaple
+  # you would do something like this:
+  #
+  #   # note this is necessary in order to segregate usage of cache between various providers regions and accounts.
+  #   # if you are using one account/region/etc only, you still must set it. 'default' will do.
+  #   Fog::Cache.sandbox_name = "prod-emea-eu-west-1"
+  #
+  #   s = security_groups.sample; s.name # => "default"
+  #   s.cache.dump # => 2371
+  #
+  # Now it is on disk:
+  #
+  # shai@adsk-lappy ~   % tree ~/.fog-cache/prod-emea-eu-west-1/
+  #
+  #   /Users/shai/.fog-cache/prod-emea-eu-west-1/
+  #     └── fog_compute_aws_real
+  #       └── fog_compute_aws_securitygroup
+  #        ├── default-90928073d9d5d9b4e7545e88aee7ec4f.yml
+  #
+  # You can do the same with a SecurityGroup, Instances, Elbs, etc.
+  #
+  # Note that when loading cache from disk, you need to pass the appropriate model class, and service associated with it.
+  # +Service+ is passed in is so that the service/connection details can be loaded into the loaded instances so they can be re-queried, etc.
+  # +Model+ is passed in so we can find the cache data associated to that model in the sandbox of cache this session is using:
+  # Will try to load all resources associated to those. If you had 1 yml file, or 100, it would load whatever it could find.
+  # As such, the normal usage of dumping would be do it on a collection:
+  #
+  #   load_balancers.each {|elb| elb.cache.dump }
+  #
+  # In order to load the cache into a different session with nothing but the service set up, use like so:
+  # As mentioned, will load all resources associated to the +model_klass+ and +service+ passed in.
+  #
+  #   instances = Fog::Cache.load(Fog::Compute::AWS::Server, compute)
+  #   instances.first.id # => "i-0569a70ae6f47d229"
+  #
+  # Note that if there is no cache located for the +model+ class and +service+ passed to `Fog::Cache.load`
+  # you will get an exception you can handle (for example, to load the resources for the fisrt time):
+  #
+  #   Fog::Cache.expire_cache!(Fog::Compute::AWS::SecurityGroup, compute)
+  #   # ... now there is no SecurityGroup cache data. So, if you tried to load it, you would get an exception:
+  #
+  #   Fog::Cache.load(Fog::Compute::AWS::SecurityGroup, compute)
+  #     rescue Fog::Cache::CacheNotFound => e
+  #       puts "could not find any cache data for security groups on #{compute}"
+  #       get_resources_and_dump
+  #
+  # == Extending cache backends
+  #
+  # Currently this is on-disk using yml. If need be, this could be extended to other cache backends:
+  #
+  # Find references of yaml in this file, split out to strategy objects/diff backends etc.
+  class Cache
+
+    # cache associated not found
+    class CacheNotFound < StandardError; end
+
+    # cache directory problem
+    class CacheDir < StandardError; end
+
+    # where different caches per +service+ api keys, regions etc, are stored
+    # see the +sandbox=+ method.
+    SANDBOX_PREFIX = "~/.fog-cache"
+
+    # when a resource is used such as `server.cache.dump` the model klass is passed in
+    # so that it can be identified from a different session.
+    attr_reader :model
+
+    # Loads cache associated to the +model_klass+ and +service+ into memory.
+    #
+    # If no cache is found, it will raise an error for handling:
+    #
+    #   rescue Fog::Cache::CacheNotFound
+    #     set_initial_cache
+    #
+    def self.load(model_klass, service)
+      cache_files = Dir.glob("#{namespace(model_klass, service)}/*")
+
+      raise CacheNotFound if cache_files.empty?
+
+      loaded = cache_files.map do |path|
+        model_klass = Object.const_get(load_cache(path)[:model_klass])
+        model_klass.new(load_cache(path)[:attrs])
+      end
+
+      collection_klass = load_cache(cache_files.sample)[:collection_klass] &&
+                         Object.const_get(load_cache(cache_files.sample)[:collection_klass])
+
+      loaded.each do |i|
+        # See https://github.com/fog/fog-aws/issues/354#issuecomment-286789702
+        i.collection = collection_klass.new(:service => service) if collection_klass
+        i.instance_variable_set(:@service, service)
+      end
+
+      # uniqe-ify based on the total of attributes. duplicate cache can exist due to
+      # `model#identity` not being unique. but if all attributes match, they are unique
+      # and shouldn't be loaded again.
+      uniq_loaded = loaded.uniq { |i| i.attributes }
+      if uniq_loaded.size != loaded.size
+        Fog::Logger.warning("Found duplicate items in the cache. Expire all & refresh cache soon.")
+      end
+
+      uniq_loaded
+    end
+
+    # creates on-disk cache of this specific +model_klass+ and +@service+
+    def self.create_namespace(model_klass, service)
+      FileUtils.mkdir_p(self.namespace(model_klass, service))
+    end
+
+    # Expires cache - this does not expire all cache associated.
+    # Instead, this will remove all on-disk cache of this specific +model_klass+ and and +@service+
+    def self.expire_cache!(model_klass, service)
+      FileUtils.rm_rf(namespace(model_klass, service))
+    end
+
+    # loads yml cache from path on disk
+    def self.load_cache(path)
+      @cache ||= {}
+      return @cache[path] if @cache[path]
+      @cache[path] = YAML.load(File.read(path))
+    end
+
+    def self.sandbox
+      @sandbox = if @sandbox_name
+                   safe_path(File.expand_path("#{SANDBOX_PREFIX}/#{@sandbox_name}"))
+                 end || (raise CacheDir.new("Must set an explicit sandbox for this cache. Example: 'serviceX-regionY'"))
+    end
+
+    def self.sandbox_name=(name)
+      @sandbox_name = name
+    end
+
+    def self.sandbox_name
+      @sandbox_name
+    end
+
+
+    # The path/namespace where the cache is stored for a specific +model_klass+ and +@service+.
+    def self.namespace(model_klass, service)
+      name = File.join(self.sandbox, service.class.to_s, model_klass.to_s)
+      name = safe_path(name)
+    end
+
+    def self.safe_path(klass)
+      klass.to_s.gsub("::", "_").downcase
+    end
+
+    def initialize(model)
+      @model = model
+    end
+
+    # Dump a Fog::Model resource. Every fog model/instance now has a +cache+ method/object injected in.
+    # as such you can use the #dump method to save the attributes and metadata of that instance as cache
+    # which can be re-used in some other session.
+    def dump
+      if !File.exist?(self.class.namespace(model.class, model.service))
+        self.class.create_namespace(model.class, model.service)
+      end
+
+      data = { :identity => model.identity,
+                     :model_klass => model.class.to_s,
+                     :collection_klass => model.collection && model.collection.class.to_s,
+                     :attrs => model.attributes }
+
+      File.open(dump_to, "w") { |f| f.write(YAML.dump(data)) }
+    end
+
+    # the location of where to save this fog model/instance to.
+    def dump_to
+      # some fog models have an identity field that is duplicate.
+      # duplicate identities can mean the cache for that already exists.
+      # this means cache duplication is possible.
+      #
+      # see "dumping two models that have duplicate identity" test case.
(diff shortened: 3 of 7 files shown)


shaiguitar:
@fog Comments? It's POC atm that can use refining but it needs some feedback to get there.

Gemfile
```diff
@@ -1,4 +1,4 @@
 source 'https://rubygems.org'
-
+gem 'rspec-core'
```
icco: This should probably be a development dependency in the gemspec?

lib/fog/core/cache.rb
```diff
@@ -0,0 +1,89 @@
...
+
+    def self.cache_data(path)
+      @cache_data ||= {}
+      return @cache_data[path] if @cache_data[path]
+      @cache_data[path] = YAML.load(File.read(path))
+    end
+
+    def self.sandbox(model_klass, service)
+      if service.nil?
+        name = "/tmp/#{safe_path(model_klass)}"
```
icco: We should probably use something like https://ruby-doc.org/stdlib-2.2.1/libdoc/tmpdir/rdoc/Dir.html#method-c-mktmpdir because Fog is used in all sorts of places, and /tmp is not always the tmpdir

lib/fog/core/cache.rb
```diff
@@ -0,0 +1,89 @@
...
+      else
+
+        # XXXX
+        # TODO ... how to get an effective service identifier?
+        # XXXX
+
+        service_identifier = service.region
```
icco: Why bother with this instead of just the model_klass?

lib/fog/core/cache.rb
```diff
@@ -0,0 +1,89 @@
...
+        # XXXX
+
+        name = "/tmp/#{safe_path(service.class)}/#{service_identifier}/#{safe_path(model_klass)}"
+      end
+      FileUtils.mkdir_p(name)
+      name
+    end
+
+    def self.safe_path(klass)
+      klass.to_s.gsub('::', '_')
```
icco: It might be worthwhile downcasing this string because some file systems are case insensitive.

spec/core/cache_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+      end
+    end
+  end
+end
+
+describe Fog::Cache do
+  it "will raise if no cache data found" do
+    Fog.mock!
+    service = Fog::SubFogTestService.new(:region => "us-east-1")
+    Fog::Cache.expire_cache!(Fog::SubFogTestModel, service)
```
icco: Question: Why do we care if there is no cache and we expire it? Or could their be an `expire_cache` method that doesn't raise?

icco:
cc @fog/core

lib/fog/core/cache.rb
shaiguitar: The service identifier (which needs refining as to how to identify it across various Fog constructs) needs to be embedded into the cache identifier:

If you use different region or AWS account and the cache is not expired, when you load via cache it can assume incorrectly that any model klass caches (for example, an AWS Server class) belongs to that service even though it belongs to a different one.

lib/fog/core/cache.rb
shaiguitar: This is a spike/POC. That's the plan, if this seems to be the way to go. Thanks for the reminder.

spec/core/cache_spec.rb
shaiguitar: Expire cache itself doesn't raise. The  `Fog::Cache::CacheNotFound` exception raises if you try to load cache, and none is found.

shaiguitar:
@fog @icco @geemus updated cache PR. At the moment the service identifier is set to an api key if it exists, because that is what suits my needs at the moment. This satisfies using caches w/out them stepping over each other with multiple aws accounts, user keys and multiple regions within them. 

I added other fixes from @icco's feedback as well. Please take a look when you get some time. Thanks!

spec/spec_helper.rb
```diff
@@ -17,3 +17,7 @@
 Dir["spec/fake_app/**/*.rb"].each do |file|
   require File.join(File.dirname(__FILE__), "..", file)
 end
+
+RSpec.configure do |config|
```
icco: You're going to need to require rspec.

lib/fog/core/cache.rb
```diff
@@ -0,0 +1,130 @@
...
+      return @cache_data[path] if @cache_data[path]
+      @cache_data[path] = YAML.load(File.read(path))
+    end
+
+    #
+    # The path/namespace where the cache is stored
+    #
+    def self.sandbox(model_klass, service)
+      service_klass = service.class
+      name = "#{Dir.tmpdir}/#{service_klass}/#{service_identifier(service)}/#{model_klass}"
```
icco: This is better done with `File.join`

lib/fog/core/cache.rb
```diff
@@ -0,0 +1,130 @@
...
+    #     set_initial_cache
+    #
+    def self.load(model_klass, service)
+      cache_files = Dir.glob("#{sandbox(model_klass, service)}/*")
+
+      if cache_files.empty?
+        raise CacheNotFound
+      end
+
+      new_instances = cache_files.map { |path|
```
icco: If you run `rubocop -a lib/fog/core/cache.rb` it'll fix this for you, but this would read better as:

```
new_instances = cache_files.map do |path|
  model_klass = cache_data(path)[:model_klass]
  model_klass.new(cache_data(path)[:attrs])
end
```

spec/spec_helper.rb
shaiguitar: What is currently here seems to work for me:

```
shai@comp ~/fog-core [add_file_yml_caching_to_models] $ be rspec spec/core/cache_spec.rb
....

Finished in 0.0061 seconds (files took 0.13397 seconds to load)
4 examples, 0 failures

Run options: --seed 11721

# Running:



Finished in 0.001934s, 0.0000 runs/s, 0.0000 assertions/s.

0 runs, 0 assertions, 0 failures, 0 errors, 0 skips
```

Do you have merge rights?

shaiguitar: Fixed other issues you laid out previously as well, thanks @icco 

What needs to be done at this point to get this released - @geemus any other feedback?

icco: The concern is your tests are failing on Travis. Try running rake and make sure that works.

geemus:
Code looks pretty reasonable so far, thanks @icco for helping drive that forward.

@shaiguitar would love to see more detailed usage docs, so I can better understand what the interface looks like and how it is expected to work (plus we'll want this anyway). Thanks!

shaiguitar:
@geemus Any guideline/styleguides for the docs?

geemus:
@shaiguitar sure, I guess something a bit more tutorial-esque would be helpful. ie instead of individual calls, giving a sense of the overall flow of how you would use this. Does that make sense?

shaiguitar:
@geemus see latest commit. What's the deal with Travis? When I run the below, the additional tests pass:

```
$ be ruby -I spec -I . spec/core/cache_spec.rb
Run options: --seed 17737

# Running:

....

Finished in 0.006278s, 637.1734 runs/s, 955.7601 assertions/s.

4 runs, 6 assertions, 0 failures, 0 errors, 0 skips
```

icco:
@shaiguitar you're not running the same command as travis is. Run `bundle exec rake travis`

shaiguitar:
@icco hat tip. Okay, just rebased/force pushed and fixed the local travis rake bug I ran into. Stepping away for a bit, let me know if there's anything else here to look at. Thanks!

shaiguitar:
Took another stab at Travis and resolved few errors but I'm not sure I can get the rest fixed w/out some help. Are there other things other than that standing in the way of this PR?

icco:
It looks fine to me, I'll leave @geemus to make the final call.

plribeiro3000:
Sorry i did not get to this before. =s

First thing i saw is that your git is not properly configured in a way that github is not recognizing your work here. You might want to fix this before getting this merged. =)

I will take a look at the code in a bit.

lib/fog/core/cache.rb
```diff
@@ -0,0 +1,183 @@
...
+    end
+
+    # Expires cache - this does not expire all cache associated.
+    # Instead, this will remove all on-disk cache of this specific +model_klass+ and and +@service+
+    def self.expire_cache!(model_klass, service)
+      FileUtils.rm_rf(sandbox(model_klass, service))
+    end
+
+    # loads yml cache from path on disk
+    def self.cache_data(path)
```
plribeiro3000: I don't want to be pick here but why not a signature like `load_cache`?

lib/fog/core/cache.rb
```diff
@@ -0,0 +1,183 @@
...
+      @cache_data ||= {}
+      return @cache_data[path] if @cache_data[path]
+      @cache_data[path] = YAML.load(File.read(path))
+    end
+
+    # The path/namespace where the cache is stored for a specific +model_klass+ and +@service+.
+    def self.sandbox(model_klass, service)
```
plribeiro3000: Same thing here. I would expect this method to create a sandbox or something like that instead of returning a name.

lib/fog/core/cache.rb
```diff
@@ -0,0 +1,183 @@
...
+    # ideally fog-core services had an identifier similar to
+    # `identity` for model classes, but until then this will do.
+    def self.service_identifier(service)
+
+      region = service.region # in a specific region
+      klass = safe_path(service.class) # for this set of resources (elbs/compute/asgs etc)
+
+      key_id = if service.instance_variable_get(:@aws_access_key_id)
+                 # for a specific api key: constrain to a specific amazon account/user.
+                 service.instance_variable_get(:@aws_access_key_id)
```
plribeiro3000: Wont this work with `aws` only?

I know this is just the first version, but if we start with this wont it break for users that are already using it later if we change it?

fog-core.gemspec
```diff
@@ -21,7 +21,9 @@ Gem::Specification.new do |spec|
   spec.add_dependency("builder")
   spec.add_dependency("excon", "~> 0.49")
   spec.add_dependency("formatador", "~> 0.2")
+  spec.add_dependency("xmlrpc") if RUBY_VERSION.to_s >= "2.4"
 
+  spec.add_development_dependency("tins") if RUBY_VERSION.to_s > "2.0"
```
plribeiro3000: What does this gem do? I could not get it from the repo README. :shameonme:

plribeiro3000:
Just minor comments. =)

Again, sorry for the big delay here. Overall the code looks good. Awesome documentation and work here!

I just have some questions that are not necessarily blockers. It kind of depends on the answers.

fog-core.gemspec
shaiguitar: I'm actually not sure but under the Gemfile.lock it's under `coveralls` so I figure it was safe to add explicitly as a development dependency. This is to fix a travis error on rubies older than 2. I don't mind rm/moving this around, whatever makes Travis happy.

lib/fog/core/cache.rb
shaiguitar: First off there are no users that are already using this as this is a feature request. However in any case it will work simply with other services and not just AWS on a cache-by cache basis. It will use `"no-ident"` as the service identifier which means it will be usable, but not across multiple users, which is what the key id is used for (segregation). Read the comment for more detail.

shaiguitar:
@geemus @lanej let me know if there are any blockers and if we can get this released soon, I'd like to use this ⬇️ 🚰  (downstream?).

Thanks!

fog-core.gemspec
plribeiro3000: Thats fine then. =)

lib/fog/core/cache.rb
plribeiro3000: Ok then. Can we use a different variable name then? like `access_key_id`?
Otherwise this might work only with `aws` because each provider has a different name for the access variables.

geemus:
@shaiguitar sure, I think there was one remaining question from @plribeiro3000 around variable naming, but seems like we are getting close!

lib/fog/core/cache.rb
shaiguitar: I may have misunderstood this comment. What variable name are you referring to? The `@aws_access_key_id` instance variable is an identifier of the credential used only in cases which the service/connection is to AWS. And that is what is used, there is no `@access_key_id` variable I can see. In other service cases this simply defaults to a single `no-ident` string name. This can be changed in the future if need be.

shaiguitar: Hey @plribeiro3000 I changed the `key_id` variable name, I hope that clarifies it.

plribeiro3000: Hmmm. I guess i wasn't clear enough.

The variable name wasn't my real concern but the fact that you are accessing an instance variable from within the service that is used by `aws` only. The instance variable `@aws_access_key_id`.

I can see that the issue is that you need to get this value from within the connection but what about you get this from a method that reads from this variable that is defined from within the service itself?
This way each provider can define the return of such method without making this class aware of something that is for one provider only.

shaiguitar: This is the point of my original comment. Having an `identity` field for services that would encapsulate that would be the best way forward. But this involves setting this method for every single service provider and they all have different interfaces. Other than the fact that I'm not sure where all of these are be defined (fog-aws/fog-openstack/fog-rackspace/fog-google/fog-etc), how would you do it in a way that takes care of future providers?

shaiguitar: @plribeiro3000 I was hoping this wouldn't require a release from other service provider specific repositories, hence the embed here. Happy to get more feedback on how this should be implemented in practice.

plribeiro3000: I'm not sure how to do it because i did not fully get the context on this code and its side effects but i do believe we should not it the way it is right now.
The idea on this project is to be a shared codebase for all providers. If we do this here we will break its main purpose.
If the solution is to add a new method then i would say we should do it here without breaking the current behaviour and add a warning message so the maintainers of the providers can tackle it and add the necessary changes.

plribeiro3000: Thinking out loud here. What we define a method called `access_key_id` in the `Fog::Service` that returns nil after logging a message saying that it should be overwritten?

Then each provider could have its own definition, like:

```ruby
def access_key_id
    instance_variable_get(:@aws_access_key_id) # for aws
end
```

geemus: Something like that sounds reasonable. Though maybe we should just call that `identity` or something like that? (ie this could be how we do what @shaiguitar suggests, though it does require providers to do their own releases to get this new behavior).

shaiguitar: I thought this over last night a bit more. This need basically comes out of needing to create a unique sandbox for a given setup: ie, for a given user, a given service account (google/aws/whatever), region, etc we need to make sure the cache doesn't step over other setups. Instead of having this auto detection, it may be more desirable to push this down to a configurable. Who knows what namespacing constraints would be needed. It's possible using STS tokens or Api keys this uniqeness constraint would be determined otherwise and not just api tokens. Users could then define this constraint in their code via an api like:

```
Fog::Cache.namespace = "I-determine-my-sandbox-path"
```
 
Then the cache data associated would go to `~/.fog/I-determine-my-sandbox-path/` . This simplifies this cache class quite a bit and would actually be more robust. I'll push a commit soon.

plribeiro3000: I think that a default value is always good but as the first version this should be enough. :+1:

shaiguitar: @geemus @plribeiro3000 pushed some modifications. Appreciate the discussion here, helped clarify some things. The flow is to set this configurable and then the rest is the same. 

```
Fog::Cache.sandbox_name = "service-aws1-region-us-west-2" 
```

This gets rid of any AWS-specific dependency since it lets users define the cache namespace/uniqueness between various accounts/users/regions. Let me know if you find anything else. Looking forward to a release and iterating over this usage over time.

geemus:
@shaiguitar thanks for working through this with us. Seems good to me.

@icco @plribeiro3000 any remaining concerns? Thanks!

lib/fog/core/cache.rb
```diff
@@ -0,0 +1,210 @@
...
+      @cache ||= {}
+      return @cache[path] if @cache[path]
+      @cache[path] = YAML.load(File.read(path))
+    end
+
+    def self.sandbox
+      @sandbox = if @sandbox_name
+                   safe_path(File.expand_path("#{SANDBOX_PREFIX}/#{@sandbox_name}"))
+                 end || (raise CacheDir.new("Must set an explicit sandbox for this cache. Example: 'serviceX-regionY'"))
+    end
```
plribeiro3000: Hmmm. Not trying to be pick but why not a guard clause here? It might make the code a little better to read:

```ruby
def self.sandbox
    raise CacheDir.new("Must set an explicit sandbox for this cache. Example: 'serviceX-regionY'") unless @sandbox_name

    @sandbox = safe_path(File.expand_path("#{SANDBOX_PREFIX}/#{@sandbox_name}"))
end
```

plribeiro3000: Also, i would prefer to use the method getter for this instance variable you defined a few lines below.

lib/fog/core/cache.rb
```diff
@@ -0,0 +1,210 @@
...
+    # Instead, this will remove all on-disk cache of this specific +model_klass+ and and +@service+
+    def self.expire_cache!(model_klass, service)
+      FileUtils.rm_rf(namespace(model_klass, service))
+    end
+
+    # loads yml cache from path on disk
+    def self.load_cache(path)
+      @cache ||= {}
+      return @cache[path] if @cache[path]
+      @cache[path] = YAML.load(File.read(path))
```
plribeiro3000: I just realized this. 

Since you are saving the result on a instance variable in the class (not into the object) and every time you call this method you save more data into that instance variable we might have some performance issues here regarding memory, no?

Is the content being saved here going to be too large? Should we have concerns about this?

lib/fog/core/cache.rb
```diff
@@ -0,0 +1,210 @@
...
+    def self.sandbox_name
+      @sandbox_name
+    end
+
+
+    # The path/namespace where the cache is stored for a specific +model_klass+ and +@service+.
+    def self.namespace(model_klass, service)
+      name = File.join(self.sandbox, service.class.to_s, model_klass.to_s)
+      name = safe_path(name)
+    end
```
plribeiro3000: Maybe you should name this variable `namespace` instead of `name`?
Just a minor detail but since you are working with static methods we might have some weird issues since this is a ruby's default static method name.

lib/fog/core/cache.rb
```diff
@@ -0,0 +1,210 @@
...
+
+    def self.sandbox
+      @sandbox = if @sandbox_name
+                   safe_path(File.expand_path("#{SANDBOX_PREFIX}/#{@sandbox_name}"))
+                 end || (raise CacheDir.new("Must set an explicit sandbox for this cache. Example: 'serviceX-regionY'"))
+    end
+
+    def self.sandbox_name=(name)
+      @sandbox_name = name
+    end
```
plribeiro3000: I'm sorry but i also just realized this as well. Since this class is supposed to be able to be extended to other backend caches in the future, maybe the preffix `sandbox` in the name might not make much sense.
Is there another name you can think of? Maybe just `storage_path` so we can have a signature like `Fog::Cache.storage_path`?

lib/fog/core/model.rb
```diff
@@ -20,6 +21,10 @@ def initialize(new_attributes = {})
       merge_attributes(attribs)
     end
 
+    def cache
+      Fog::Cache.new(self)
+    end
```
plribeiro3000: What if we expose only the `dump` and `load` methods from `Cache` instead of the whole object?
Do we need everything to be available form the model itself?

plribeiro3000:
Hey @shaiguitar .

I'm sorry for the extra comments but as you've refactored the code got a little bit more clear for me to understand its usage. 😅

lib/fog/core/cache.rb
plribeiro3000: or just `Fog::Cache.path` ?!

lib/fog/core/cache.rb
shaiguitar: This is only called once, and it's stored per-path (ie, resource of a model/service pair). This should not be a concern, and it can be revisited if need be later.

lib/fog/core/cache.rb
shaiguitar: Updated.

lib/fog/core/cache.rb
shaiguitar: Updated.

lib/fog/core/cache.rb
shaiguitar: I think with `sandbox` I would be happy to leave the name `sandbox_name`. We can rename these if someone at some point adds another cache backed. There is no harm in refactoring the cache class at large if there is a need in the future since this would just invalidate the cache once and be done with it.

lib/fog/core/model.rb
shaiguitar: The only method that is available from a `Fog::Model` is via this injected Cache object, and the only instance methods that exist are `model.cache.dump` and `model.cache.dump_to`. Everything else is a Cache class method so not available via this.

shaiguitar:
I pushed some last commits. @plribeiro3000 If you'd like to fork and submit any further changes please go ahead otherwise I think the rest we should look into if and when anyone feels like adding another backend. Until then, I think this should do. This branch was tested within a short lived CLI application that uses many different apps and different model resources.

shaiguitar:
Is it worth trying to fix the ruby 1.8.7 errors on travis or is that not a blocker? What is at this point?

icco:
@shaiguitar no worries on 1.8.7, I opened #205 for that. Just need @plribeiro3000's sign off.

lib/fog/core/cache.rb
plribeiro3000: Hmm. `Fog::Cache.sandbox_name` just don't feels right.
Is not just about what it can be in the future but also that If this method is defining the namespace or path it should state that.

lib/fog/core/model.rb
plribeiro3000: Hmmm. Actually all methods are available through `.class` on the instance but let it be. =)
We can refactor this later. No need to block this PR because of this.

lib/fog/core/cache.rb
plribeiro3000: Just to be sure. What does it store exactly?
I'm still worried about this.

plribeiro3000:
I guess the only thing left is the method name now. 
After that we are good to go with this. =)

Great work here!

lib/fog/core/cache.rb
shaiguitar: This is what stores the data from the yml files being loaded in. It's memoized because it's hit a few times within `Cache.load`. For what it's worth and to address your feedback, I set this variable to nil, so memory should be freed after model initialization/loading is done.

lib/fog/core/cache.rb
shaiguitar: Updated to `namespace_prefix` to reflect it's connection to defining the `Cache.namespace` variable.

shaiguitar:
Updated, look forward to a version number.

plribeiro3000:
Thats for all your workshop here! 👍

shaiguitar:
Thanks. Hopefully it's useful to others beyond my use case.

Can we do a version bump & release?

geemus:
Done: v1.44.0

geemus:
Thanks for all discussing and working through this together!

PR #209, opened 2017-05-09
Cache fixes

shaiguitar:
Not important to merge/release at this point, as there might be some more iterations to go through. But logging this here so I don't forget.

There might be some more fixes of these sorts of things over time as usage increases.

### DIFF
--- a/lib/fog/core/cache.rb
+++ b/lib/fog/core/cache.rb
@@ -109,34 +109,73 @@ module Fog
 
       raise CacheNotFound if cache_files.empty?
 
-      attributes = cache_files.map do |path|
-        load_cache(path)[:attrs]
+      # collection_klass and model_klass should be the same across all instances
+      # choose a valid cache record from the dump to use as a sample to deterine
+      # which collection/model to instantiate.
+
+      path = cache_files.detect{ |path| valid_for_load?(path) }
+      model_klass = const_get_compat(load_cache(path)[:model_klass])
+      collection_klass = if valid_for_load?(path) && load_cache(path)[:collection_klass]
+                           const_get_compat(load_cache(path)[:collection_klass])
+                         end
+
+      loaded = cache_files.map do |path|
+          model_klass.new(load_cache(path)[:attrs]) if valid_for_load?(path)
+      end.compact
+
+      # Load collection and service so they can be reloaded/connection is set properly.
+      # See https://github.com/fog/fog-aws/issues/354#issuecomment-286789702
+      loaded.each do |i|
+        i.collection = collection_klass.new(:service => service) if collection_klass
+        i.instance_variable_set(:@service, service)
       end
+
       # uniqe-ify based on the total of attributes. duplicate cache can exist due to
       # `model#identity` not being unique. but if all attributes match, they are unique
       # and shouldn't be loaded again.
-      uniq_attributes = attributes.uniq
-      if uniq_attributes.size != attributes.size
+      uniq_loaded = uniq_w_block(loaded) { |i| i.attributes }
+      if uniq_loaded.size != loaded.size
         Fog::Logger.warning("Found duplicate items in the cache. Expire all & refresh cache soon.")
-      end
-
-      loaded = uniq_attributes.map do |attrs|
-        model_klass.new(attrs)
-      end
-
-      collection_klass = load_cache(cache_files.first)[:collection_klass] &&
-                         const_split_and_get(load_cache(cache_files.first)[:collection_klass])
-
-      loaded.each do |i|
-        # See https://github.com/fog/fog-aws/issues/354#issuecomment-286789702
-        i.collection = collection_klass.new(:service => service) if collection_klass
-        i.instance_variable_set(:@service, service)
       end
 
       # Fog models created, free memory of cached data used for creation.
       @memoized = nil
 
-      loaded
+      uniq_loaded
+    end
+
+    # :nodoc: compatability for 1.8.7 1.9.3
+    def self.const_get_compat(strklass)
+      # https://stackoverflow.com/questions/3163641/get-a-class-by-name-in-ruby
+      strklass.split('::').inject(Object) do |mod, class_name|
+        mod.const_get(class_name)
+      end
+    end
+
+    # :nodoc: compatability for 1.8.7 1.9.3
+    def self.uniq_w_block(arr)
+      ret, keys = [], []
+      arr.each do |x|
+        key = block_given? ? yield(x) : x
+        unless keys.include? key
+          ret << x
+          keys << key
+        end
+      end
+      ret
+    end
+
+    # method to determine if a path can be loaded and is valid fog cache format.
+    def self.valid_for_load?(path)
+      data = load_cache(path)
+      if data && data.is_a?(Hash)
+        if [:identity, :model_klass, :collection_klass, :attrs].all? { |k| data.keys.include?(k) }
+          return true
+        else
+          Fog::Logger.warning("Found corrupt items in the cache: #{path}. Expire all & refresh cache soon.\n\nData:#{File.read(path)}")
+          return false
+        end
+      end
     end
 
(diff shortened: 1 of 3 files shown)


lib/fog/core/cache.rb
```diff
@@ -110,9 +110,13 @@ def self.load(model_klass, service)
...
       loaded = cache_files.map do |path|
-        model_klass = Object.const_get(load_cache(path)[:model_klass])
-        model_klass.new(load_cache(path)[:attrs])
-      end
+        if load_cache(path) # empty files, bad cache data.
+          model_klass = Object.const_get(load_cache(path)[:model_klass])
+          model_klass.new(load_cache(path)[:attrs])
+        else
+          Fog::Logger.warning("Found corrupt items in the cache: #{path}. Expire all & refresh cache soon.\n\nData:#{File.read(path)}")
+        end
```
plribeiro3000: I wonder where this check should be done here or in the `load_cache` method instead.
This one is big enough already.

geemus: Sounds reasonable.

BigGillyStyle:
Just as an FYI...we've had to revert to fog-core 1.44.1 gem version. 1.44.2 breaks with the follow error:

```
/usr/local/rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/bundler-1.12.3/lib/bundler/runtime.rb:89:in rescue in block (2 levels) in require': There was an error while trying to load the gem 'fog'. (Bundler::GemRequireError) Gem Load Error is: couldn't find HOME environment -- expanding~'
Backtrace for gem load error is:
[release_directory]/vendor/bundle/ruby/2.3.0/gems/fog-core-1.44.2/lib/fog/core/cache.rb:94:in `expand_path'
```

Credit to @PashaLVWD for finding this for us.  He led me to this PR, which may be a fix?

geemus:
@BigGillyStyle Thanks for the heads up. I think this contains the fix (but also some other stuff which is still in progress). I'll see if I can't extract the relevant bit so we can cut that into a release.

geemus:
@BigGillyStyle I think it should be fixed and released in v1.44.3, sorry for the breakage!

BigGillyStyle:
@geemus No worries...I mostly wanted to just let you know about this.  I just tested out 1.44.3 and we're all good.  Awesome turnaround! 👍

shaiguitar:
Been running this fork for a while and seems to work w/out an issue. 

- Added `Fog::Cache.metadata` for a given namespace so you can create custom cache expiry policies, or whatever you'd like to do for a given cache namespace / cloud vendor account-region etc.

For example:

```
  def renew_cache?
    last_refreshed = Fog::Cache.metadata[:last_refreshed]

    if last_refreshed.nil?
      # no metadata
      return true
    end

    if (Time.now - last_refreshed) > EXPIRE_TIME
      return true
    else
      return false
    end
  end
end
``` 

- Added `Fog::Cache.clean!` as a helper method to blow away all local cache not just a specific namespace.
- Removed the `$HOME` env expand path fix that has since been merged.
- Refactored some code out of the large method into a separate check.

FWIW, locally the tests pass:

```
$ ruby -v
ruby 2.2.5p319 (2016-04-26 revision 54774) [x86_64-darwin16]

$ bundle exec rake travis

# Running:

................................................................../users/shai/.fog-cache/for-service-user-region-foo/fog_subfogtestservice_mock/fog_subfogtestmodel
/users/shai/.fog-cache/for-service-user-region-foo
..[fog][WARNING] Found duplicate items in the cache. Expire all & refresh cache soon.
..................................................................................................S..........................................................

Finished in 7.095842s, 31.7087 runs/s, 38.6142 assertions/s.

225 runs, 274 assertions, 0 failures, 0 errors, 1 skips

You have skipped tests. Run with --verbose for details.
```

geemus:
@shaiguitar there were some weird breaks on master, which I think are fixed now. Could you rebase? Hopefully that will get us back to a passing test suite here. Thanks!

shaiguitar:
@geemus rebased. looks like travis is failing on https://stackoverflow.com/questions/3163641/get-a-class-by-name-in-ruby for 1.9.3 and 1.8.7 ... Are those deprecated or should I try to patch those?

shaiguitar:
So fixed the const_get compat and running into a different 1.8.7 compat issue with `Array#sample`. Is 1.8.7 supported or is that planned to be removed from travis?

geemus:
I think if you rebase off master it should help. I fixed the const_get issue there, as well as other dependency issues (and got master passing for everything). I think those fixes are pretty good/clean (and are apt to conflict with these changes). Does that help/make sense?

shaiguitar:
It has been rebased

shaiguitar:
I thought I had rebased earlier but now I'm wondering if I screwed up the rebase - can you take a quick look at the branch's current state?

geemus:
I see my changes in the `-` portion of the diffs, but it looks like post-rebase they have all been removed (ie things like the const_split_and_get method, which I added to deal with the const issue). So maybe the merge choices were off?

shaiguitar:
Yeah at this point `git diff origin/master` points to that const get compat method that you added being removed, which ironically I added a similar one after rebasing and not including that in. Right now the only thing broken on travis is a `Array#sample` method that should be `Array#choice` 1.8.7. I could add another compat method for that if 1.8.7 is still to be around on CI. Once that specific branch issue is cleared I imagine it should go green assuming there's nothing else wrong with the rebase.

shaiguitar:
@geemus Latest compat error was 1.8.7 does not take `uniq` with a block. After this latest method addition travis passes. Unless you think moving the various compat errors from the module into core_ext or some sort makes sense, this should be ready to merge.

lib/fog/core/cache.rb
```diff
@@ -109,34 +109,81 @@ def self.load(model_klass, service)
 
       raise CacheNotFound if cache_files.empty?
 
-      attributes = cache_files.map do |path|
-        load_cache(path)[:attrs]
+      loaded = cache_files.map do |path|
+        if valid_for_load?(path)
+          model_klass = const_get_compat(load_cache(path)[:model_klass])
```
geemus: Does this need to be repeated? Shouldn't it be the same as the passed-in model_klass value that is already set in the method scope?

lib/fog/core/cache.rb
```diff
@@ -109,34 +109,81 @@ def self.load(model_klass, service)
...
+          model_klass.new(load_cache(path)[:attrs])
+        end
+      end.compact
+
+      path = pick_random(cache_files)
```
geemus: I wasn't clear on this, why does this use a random file?

spec/core/cache_spec.rb
```diff
@@ -35,9 +35,63 @@ def initialize(opts = {})
     example_cache = File.expand_path(Fog::Cache.namespace(Fog::SubFogTestModel, @service)).downcase
     expected_namespace = File.expand_path("~/.fog-cache/for-service-user-region-foo").downcase
 
+    puts example_cache
```
geemus: Could you remove these `puts`? I think I pulled them on master, as I presume they were just accidentally left behind?

spec/core/cache_spec.rb
```diff
@@ -101,7 +173,7 @@ def initialize(opts = {})
 
     id = SecureRandom.hex
 
-    # security groups on aws for eg can have the same identity group name 'default'.
+    # security gruops on aws for eg can have the same identity group name 'default'.
```
geemus: I think I fixed this typo in my master also, looks like that change also was dropped.

shaiguitar:
@geemus Understood. Rebase went off due to many commits I hadn't squashed so I opt'ed to skip some and just fill in the differences manually, it was my mistake. 

I fixed the recent comments - removed random sampling by simply finding the first cache entry that is valid and the puts/comment typo. I also moved model/collection detection out of the loop. Let me know if there's anything left outstanding.

geemus:
Thanks for working through that with me, looks good.

PR #217, opened 2017-09-26
Update associations on reload

### LINKED ISSUE #216: Model::reload does not update associations
When calling `.reload` on a model `Foo`, the attributes are updated, but the associations remain the same, even if `Foo.get` `load`ed new elements in an association.

joshuabach:
Closes #216.

I'm not sure if what I did in `lib/fog/core/attributes.rb:131` is enough. Do I need to take special care of association aliases?

### DIFF
--- a/lib/fog/core/attributes.rb
+++ b/lib/fog/core/attributes.rb
@@ -127,6 +127,11 @@ module Fog
         self
       end
 
+      def merge_associations(new_associations = {})
+        associations.update(new_associations)
+        self
+      end
+
       # Returns true if a remote resource has been assigned an
       # identity and we can assume it has been persisted.
       #
--- a/lib/fog/core/model.rb
+++ b/lib/fog/core/model.rb
@@ -54,6 +54,7 @@ module Fog
 
       new_attributes = data.attributes
       merge_attributes(new_attributes)
+      merge_associations(data.associations)
       self
     end
 


plribeiro3000:
Hmmm. I'm not sure if this is gonna work the way you intend. 
Let me check the code to remember some details and then i will get back here.

plribeiro3000:
Ok. Just got it back in my head. =)
So the way it works exactly is a bit different from the attributes.
When initializing a new model, it will set the attributes related to an association on this association array.
The way it is saved on this array might differ depending on the provider.
If the provider does return just the identifier it will save just the plain identifier value and when retrieving the association it will load a new object.
It the provider does return the association object out of the box it will save that object on the associations array.
When initializing an object it does set its association and when reloading it should do it out of the box [here](https://github.com/fog/fog-core/blob/3895821a2c5087e95f5ea726c8071c62e78dc4ee/lib/fog/core/model.rb#L56).
Checking the code of `merge_attributes` and `reload` i realized it does not actually `reload` but just overwrite every single value.
The problem might be that the provider you are using is not returning all the data of the object and thus it is using the old data.
If i'm not wrong the solution is to chance `reload` to actually wipe any data before setting the new one but that might need input from other maintainers and other providers.

Meanwhile, can you provide mode info, like provider you are using and if it does return all the data right away or if you need extra calls?

cc/ @geemus

geemus:
Yeah, I had wondered if something like nilling and waiting for lazy load might be the right way to go. But I am a bit concerned about how universally this can be done.

@joshuabach more details of your specific situation and usage would definitely help us to understand how best to approach this. Thanks!

joshuabach:
My provider is a self built one (no public repo yet, so no link, sorry). I'm working on a fog provider (and Foreman plugin, at that) for IBM PowerVM/HMC [1]. I'll just quickly summarise what I intend to do:

My `Server` model `has_many_identities` of `Interface`, interfaces are `id`ed by a `:mac` attribute.
`Servers:all` looks sort of like this (much simplified):

```ruby
def all(filters)
  load(service.get_servers
            .select {|s| s.matches filters } # Pseudo function
            .map    {|s| s.merge(
                            :interfaces =>
                            service.get_ifaces(s.id)
                              .map{|iface| iface[:mac_addr]})})
end

def get(id)
  all(:id => id).first
end
```
So what is passed to `load`, looks like this:
```ruby
[
  {
    # Attributes...
    :interfaces => [
      "some:mac:address",
      # ...
    ]
  }
]
```

So `service.servers.get(id).interfaces` is now an association instance.
The problem with HMC is that, when I `services.servers.create(...)` a new VM,
the mac addresses are not known, but assigned by a DHCP server on first boot, so it essentially comes down to:

```ruby
irb> s = service.servers.create(...)
irb> s.start
irb> s.stop 
# Now the MAC addresses have been assigned
irb> s.reload.interfaces
=> nil
irb> service.servers.get(s.id).interfaces
=> <Fog::Association [...]>
```

joshuabach:
I' just realised that my `s.interfaces` returns a `Fog::Association` of `Interfaces` collections,
which each contain one `Interface` (model). Shouldn't It be a `Fog::Association` of `Interface` models?

Maybe I just understood something wrong, and the fault is mine?

plribeiro3000:
OK. So this

```ruby
irb> s = service.servers.create(...)
irb> s.start
irb> s.stop 
# Now the MAC addresses have been assigned
irb> s2 = service.servers.get(s.id)
irb> s2.interfaces
=> <Fog::Association [...]>
irb> s2.associations
=> {:interfaces=>[...]}
irb> s.reload.interfaces
=> nil
irb > s.associations
=> nil
```

gives me a hint. =)

When you first load the object it does have the association but does not when you reload. Somehow both methods are working in a different way (which they should not).

Let me check the code and get back here with what i discover.

plribeiro3000:
One stupid question here. Just to be sure. heheh
In your code you are doing a get on the collection passing `#id` but `#reload` does use the `#identity` attribute.

Did you define the `identity` in your model?

joshuabach:
Yes I did, like this:
```ruby
class Server
  identity :id
  ...
``` 
For sake of completeness, `#id` is a 2-element Array, set in the constructor:
```ruby
...
def initialize(attributes={})
  if attributes.has_key?(:name) and attributes.has_key?(:system_name)
    self.id = [attributes[:name],
               attributes[:system_name]]
  end
  super
end
...
```

plribeiro3000:
Hmmm. Can you confirm the value of `id` after creating the object, and after the reload?

joshuabach:
Yes.
```ruby
irb> s = service.servers.create(:system_name => 'foo', :name => 'bar')
irb> s.id
=> ["blub", "Server-8284-22A-SN21602BV"]
irb> s.reload.id
=> ["blub", "Server-8284-22A-SN21602BV"]
irb> service.servers.get(['bar', 'foo']).id
=> ["blub", "Server-8284-22A-SN21602BV"]
```
I can also confirm that my `Servers::all` actually behaves the same both on `reload`ing and on `get`ing, i.e. that the exact same `Array` of `Hash`es is passed to `load`.

plribeiro3000:
Can you check the response of your provider when you are creating a new object and when you are just getting it? Like the actual response?

joshuabach:
What exactly do you mean by _response_?

The `Server` object _is_  in fact incomplete after creation, as the MAC addresses / interfaces are only known after the first boot in HMC, only a new `get` after that returns all attributes / associations (and reloading an existing server should result in the same object as getting the server with the same `#id`, if I understand that correctly, as `reload` internaly uses `get`).

So essentially (IMO), `s.reload` should return an object equivalent to `Servers.get(s.identity)` (same attributes, same associations).

plribeiro3000:
I agree it should behave in the same way. Thats why i want to check the response of both scenarios.

Can you provide the http response of both cases plz? I want to double check it.

joshuabach:
There is no HTTP response. HMC is a command line interface, which I invoke over SSH (using `net-ssh`). A single `get` (by which I mean the function`Servers::get`, not the HTTP method `GET`, if that caused the confusion) requires two HMC commands, one to get the attributes (invoked by `service.get_servers` in the above example) and one to get the mac addresses (`service.get_ifaces(s.id)`).

The result  of the first looks like this (comma separated key-value pairs): 
```
name=foo,lpar_id=13,lpar_env=aixlinux,state=Not Activated,resource_config=1,os_version=Unknown,logical_serial_num=21602BVD,default_profile=default,curr_profile=default,work_group_id=none,shared_proc_pool_util_auth=0,allow_perf_collection=0,power_ctrl_lpar_ids=none,boot_mode=norm,lpar_keylock=norm,auto_start=0,redundant_err_path_reporting=0,rmc_state=inactive,rmc_ipaddr=,time_ref=0,lpar_avail_priority=127,desired_lpar_proc_compat_mode=default,curr_lpar_proc_compat_mode=POWER8,suspend_capable=0,remote_restart_capable=0,simplified_remote_restart_capable=0,sync_curr_profile=0,affinity_group_id=none,vtpm_enabled=0,powervm_mgmt_capable=0
```
The response returned over SSH in the `get_ifaces` method looks like this (tuples of lpar_id, name, vlan_id, mac_addr:
```
13,foo,1420,1A14DC434702
13,bar,343,1A14DC434703
```
And _yes_, after the first start of the VM, both results are the same in both cases.

geemus:
How are you setting up the association?

plribeiro3000:
Ok. Let me dive into the code and try to see if i can catch something.

geemus:
@plribeiro3000 thanks, I'm not so familiar with the association stuff, so I appreciate the help!

joshuabach:
@geemus As seen in the first snippet I postest, by passing the key `:interfaces:` to `load` in `Servers:all`.
The `Interfaces` collection looks like this:
```ruby
class Interfaces < Collection
  model Fog::Compute::PowerVM::Models::Interface
  def get(mac)
    load([{:mac => mac}]).first 
  end
end

class Interface < Fog::Model
  identity :mac 
  has_one_identity :server, :servers
end

```

plribeiro3000:
Ok, so your problem might be here:

https://github.com/fog/fog-core/blob/13979be5237d7cd7d9a99f2dccbe2fe138ff69aa/lib/fog/core/associations/many_identities.rb#L20-L28

Can you try to debug it? 

Since this is returning `nil` i believe the error might be in [this line](https://github.com/fog/fog-core/blob/master/lib/fog/core/associations/many_identities.rb#L23) which means the it is an empty array so we need to check that  instead of just `nil?`.

I'm still diving into it to see if i can figure out more about it.

plribeiro3000:
Ok, going further i believe

https://github.com/fog/fog-core/blob/13979be5237d7cd7d9a99f2dccbe2fe138ff69aa/lib/fog/core/attributes.rb#L72-L74

is empty after reloading.

Since the code at `reload` is pretty straightforward i believe it might be in the setter.

So could you debug this:

https://github.com/fog/fog-core/blob/13979be5237d7cd7d9a99f2dccbe2fe138ff69aa/lib/fog/core/associations/many_identities.rb#L10-L16 

and paste whats exactly happening in there? It might be that the `Array(new_#{name})` is returning nil.

joshuabach:
(All traces when invoking `reload` on a newly `create`ed `Server`)

First:
```
irb(main):025:0> foo.interfaces

[17, 26] in /home/joshua/repos/by-owner/fog/fog-core/lib/fog/core/associations/many_identities.rb
   17:         EOS
   18:       end
   19:
   20:       def create_getter
   21:         model.class_eval <<-EOS, __FILE__, __LINE__
=> 22:           def #{name}
   23:             byebug
   24:             return [] if associations[:#{name}].nil?
   25:             data = Array(associations[:#{name}]).map do |association|
   26:               service.send(self.class.associations[:#{name}]).get(association)
(byebug) associations
{}
(byebug) associations.nil?
false
```

Second:
```
[7, 16] in /home/joshua/repos/by-owner/fog/fog-core/lib/fog/core/associations/many_identities.rb
    7:     # The association models will be loaded based on the collection of ids.
    8:     class ManyIdentities < Default
    9:       def create_setter
   10:         model.class_eval <<-EOS, __FILE__, __LINE__
   11:           def #{name}=(new_#{name})
   12:             byebug
=> 13:             associations[:#{name}] = Array(new_#{name}).map do |association|
   14:                                        association.respond_to?(:identity) ? association.identity : association
   15:                                      end
   16:           end
(byebug) associations
{}
(byebug) Array(new_interfaces)
["1A14D6F53702", "1A14D6F53703"]
(byebug) next

[122, 131] in /home/joshua/repos/by-owner/fog/fog-core/lib/fog/core/attributes.rb
   122:             send("#{key}=", value)
   123:           else
   124:             attributes[key] = value
   125:           end
   126:         end
=> 127:         self
   128:       end
   129:
   130:       # Returns true if a remote resource has been assigned an
   131:       # identity and we can assume it has been persisted.
(byebug) self
  <Fog::Compute::PowerVM::Models::Server
    id=["bar", "foo"],
    system_name="foo",
    name="bar",
    lpar_id="18",
    lpar_env="aixlinux",
    state="Not Activated",
    resource_config="1",
    os_version="Unknown",
    logical_serial_num="21602BVI",
    default_profile="default",
    curr_profile="default",
    work_group_id="none",
    shared_proc_pool_util_auth="0",
    allow_perf_collection="0",
    power_ctrl_lpar_ids="none",
    boot_mode="norm",
    lpar_keylock="norm",
    auto_start="0",
    redundant_err_path_reporting="0",
    rmc_state="inactive",
    rmc_ipaddr=nil,
    time_ref="0",
    lpar_avail_priority="127",
    desired_lpar_proc_compat_mode="default",
    curr_lpar_proc_compat_mode="POWER8",
    suspend_capable="0",
    remote_restart_capable="0",
    simplified_remote_restart_capable="0",
    sync_curr_profile="0",
    affinity_group_id="none",
    vtpm_enabled="0",
    powervm_mgmt_capable="0\n"
  >
(byebug) associations
{:interfaces=>["1A14D6F53702", "1A14D6F53703"]}
```

I also debugged lib/fog/core/model.rb:54 (`#reload`)
```
[51, 60] in /home/joshua/repos/by-owner/fog/fog-core/lib/fog/core/model.rb
   51:       end
   52:
   53:       return unless data
   54:       byebug
   55:
=> 56:       new_attributes = data.attributes
   57:       merge_attributes(new_attributes)
   58:       self
   59:     end
   60:
(byebug) data.associations
{:interfaces=>["1A14D6B12702", "1A14D6B12703"]}
(byebug) self.associations
{}
(byebug) n

[52, 61] in /home/joshua/repos/by-owner/fog/fog-core/lib/fog/core/model.rb
   52:
   53:       return unless data
   54:       byebug
   55:
   56:       new_attributes = data.attributes
=> 57:       merge_attributes(new_attributes)
   58:       self
   59:     end
   60:
   61:     def to_json(_options = {})
(byebug) n

[53, 62] in /home/joshua/repos/by-owner/fog/fog-core/lib/fog/core/model.rb
   53:       return unless data
   54:       byebug
   55:
   56:       new_attributes = data.attributes
   57:       merge_attributes(new_attributes)
=> 58:       self
   59:     end
   60:
   61:     def to_json(_options = {})
   62:       Fog::JSON.encode(attributes)
(byebug) self.attributes
{:id=>["foo", "bar"], :system_name=>"bar", :name=>"foo", :lpar_id=>"14", :lpar_env=>"aixlinux", :state=>"Running", :resource_config=>"1", :os_version=>"Unknown", :logical_serial_num=>"21602BVE", :default_profile=>"default", :curr_profile=>"default", :work_group_id=>"none", :shared_proc_pool_util_auth=>"0", :allow_perf_collection=>"0", :power_ctrl_lpar_ids=>"none", :boot_mode=>"norm", :lpar_keylock=>"norm", :auto_start=>"0", :redundant_err_path_reporting=>"0", :rmc_state=>"inactive", :rmc_ipaddr=>nil, :time_ref=>"0", :lpar_avail_priority=>"127", :desired_lpar_proc_compat_mode=>"default", :curr_lpar_proc_compat_mode=>"POWER8", :suspend_capable=>"0", :remote_restart_capable=>"0", :simplified_remote_restart_capable=>"0", :sync_curr_profile=>"0", :affinity_group_id=>"none", :vtpm_enabled=>"0", :powervm_mgmt_capable=>"0\n"}
(byebug) self.associations
{}
```
So as far as I understand it, the updated object is correctly retrieved from the provider (as `reload` simply uses `get` to retrive the new object, which we verified that it works). But then, only the `#attributes` hash is transfered from `data` to `self`, and the `#associations` are lost. So the bug is in the `#reload` function, when transferring the data from the temporary object to `self`.

plribeiro3000:
Ok. You are right. I will send a fix right now. Thank you for bearing with me on this!

plribeiro3000:
Could you check the branch [fix-reload-for-associations](https://github.com/fog/fog-core/tree/fix-reload-for-associations) to see if it is fixed please?

joshuabach:
Will do so on monday!

joshuabach:
Just tried your branch, and it works! Both the attributes and the associations are now correctly updated on `#reload`.

I wonder if there is a way to do this in a manner that is more robust to future changes, e.g. when new members are added to `Model`, apart from associations and attributes. A shallow copy with `Object#clone` or `Object#dup` would probably be enough (as the `result_object` is immediately discarded), but then again, these functions also just return a new object. I don't think there is something in ruby that does "copy object `X` into the existing object `Y`", or sth. 

But well, I am happy with your fix. Thanks!

plribeiro3000:
Yeap. Ruby does not have a way for the object to copy a new object into self.
What we can do is to make this code a little bit cleaner (and clear) about what it does.
The fact that the result object was being called `data` was taking my attention out of it.

since it worked, i will do some more cleaning on it and then submit a PR.

Thanks again for working with me on this. 👍

geemus:
Going to go ahead and close this for clarity (further discussion over on the branch). Just let me know if I am incorrect in doing so. Thanks!

PR #218, opened 2017-10-02
fix association reload

plribeiro3000:
What should be the expect behavior of reload when it fails?

Because the current one is a little bit off the road IMHO.

If it works returns `self`, otherwise `nil`.

Maybe we should return `self` if it works and raise an exception if fails? Or always return self?
WDYT?

cc/ @joshuabach @lanej @icco @Temikus


lib/fog/core/associations/many_identities.rb
```diff
@@ -19,7 +19,7 @@ def #{name}=(new_#{name})
       def create_getter
         model.class_eval <<-EOS, __FILE__, __LINE__
           def #{name}
-            return [] if associations[:#{name}].nil?
+            return [] if associations.empty? || associations[:#{name}].nil?
```
geemus: I think I understand the other changes, but I'm less sure about this one. Could you elaborate?

icco:
I guess I don't understand in this scenario how I detect no associations versus network problems? Like when do I retry? Or am I missing something?

lib/fog/core/associations/many_identities.rb
plribeiro3000: Hmmmm. I guess you are right. I was trying to solve another issue, but this code changes nothing.

For some reason it was returning nil when the association did not exist when it should return an empty array because it is a collection. Will remove this change.

plribeiro3000:
@icco Yeap.

I suggest we change the current implementation to raise a custom exception in case the retrieval of data goes wrong and in any other case always return self so chaining call won't break for different results. What do you think? Sounds good?

lanej:
@plribeiro3000 I'd prefer an exception.  End users would probably like to know why it failed.

plribeiro3000:
@lanej Agreed. Once this fix got merged i will send another PR changing the behaviour.
This way i can easily spot the difference in the future.

geemus:
Thanks all!

PR #219, opened 2017-10-24
Drop Ruby<2 support

tbrisker:
According to README, fog-core does not support Ruby 1.9. Users
requiring it should use the fog gem.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: ruby
 
 rvm:
+  - 2.0
   - 2.1
   - 2.2
   - 2.3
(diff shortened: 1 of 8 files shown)


geemus:
I asked this in the other thread too, but just to clarify. What happens if you are on 1.9 and try to require fog after this change? Does it still work, but you just get the latest version that allows <2.0? Or does it just fail? (apologies wrt my inexperience with this).

plribeiro3000:
You will get the last version without this constraint which would be `1.45.0`.

geemus:
Thanks for the clarification. Sounds reasonable enough (I was mostly worried about a hard stop or things suddenly breaking).

geemus:
Thanks!

plribeiro3000:
I believe we should probably go the `2.0.0` route now.

PR #220, opened 2017-10-27
Updating net-ssh parameter names

JustMcAfee:
In response to https://github.com/net-ssh/net-ssh/commit/c3fbcdc882f6763fa718f6adbd9d1d60eaf04b62

### DIFF
--- a/lib/fog/core/scp.rb
+++ b/lib/fog/core/scp.rb
@@ -64,7 +64,7 @@ module Fog
 
         @address  = address
         @username = username
-        @options  = { :paranoid => false }.merge(options)
+        @options  = { :paranoid => false, :verify_host_key => false }.merge(options)
       end
 
       def upload(local_path, remote_path, upload_options = {}, &block)
--- a/lib/fog/core/ssh.rb
+++ b/lib/fog/core/ssh.rb
@@ -58,7 +58,7 @@ module Fog
 
         @address  = address
         @username = username
-        @options  = { :paranoid => false }.merge(options)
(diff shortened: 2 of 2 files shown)


plribeiro3000:
Thanks @JustMcAfee.

To change that we need to either lock `net-ssh` version to the newest one or
check for both and allow people to use older versions of `net-ssh` (just in case the are locked)

I believe we should go the latter and set both parameters. WDYT?

JustMcAfee:
My understanding was that net::ssh is not included as an explicit dependency to fog by intent (I am not aware of the reasons for this), so locking it to a minimum version seems unlikely.  The purpose of this PR is to suppress the warning messages the latest net:ssh includes if :paranoid is at all specified, even if it is alongside :verify_ssh_key.  I think if the plan is to continue to not specify net::ssh (which seems incorrect to me personally, as it's explicitly called), then I can specify both.  Ideally, though, we'd re-add the dependency.

plribeiro3000:
@JustMcAfee Sure. This decision was made a while ago to keep support of old ruby versions.
Even that it might not be the case now, we need to specify both keys in this change.

We can revisit this `net-ssh` once we are done with the ruby < 2 odyssey.

JustMcAfee:
@plribeiro3000 Understood and updated accordingly

plribeiro3000:
👍

geemus:
The dependency was omitted because some people use fog for purely storage (or similar), where it doesn't make any sense. As we split out individual providers it becomes less of an issue, but that was why we did it initially anyway. I agree that it's a bit odd and unfortunate, but there still is at least some reason behind it.

Thanks!

Issue #203, opened 2017-03-23
mime/types gem needed in the core


gildub:
This is not a new issue [1],

meanwhile it definitely seems to make sense to have fog-core to tackle it and include the gem as discussed in [2].

[1] https://github.com/fog/fog-core/issues/177
[2] https://github.com/fog/fog-openstack/pull/196

icco:
As stated in #177, until Fog removes 1.9.3 as a supported ruby, we cannot.

 > mime-types was unbundled from fog gem to prevent ruby <2.0 compatibility issues, see: fog/fog#3775

The fog issue to follow for removing 1.9.3 support is fog/fog#3872

I'll leave this open as it is something we want to eventually do once fog 2.0 is released.

gildub:
@icco 

Understand, we've to keep it that way until Ruby 1.9.3 is removed.
Hopefully this is about to happen soon enough.

syguer:
It has begun https://github.com/fog/fog/pull/3974

plribeiro3000:
@gildub Do you have some spare time to send a PR adding it?

Thanks!

gildub:
@plribeiro3000, a PR to do what? Sorry but I'm losing track here, PR#221 has addressed the missing dependency isn't ?

plribeiro3000:
Yeap. Not sure what I was thinking. Heheh

Thanks @gildub!

Issue #177, opened 2016-03-07
`mime-types` gem is not added in the gemspec


NimishaS:
`mime-types` gem is required here https://github.com/fog/fog-core/blob/master/lib/fog/storage.rb#L11, but it's not added in the `gemspec` dependencies. 
This is causing Travis failure for the gems which are using `fog`.

Temikus:
@NimishaS `mime-types` was unbundled from `fog` gem to prevent ruby <2.0 compatibility issues, see: https://github.com/fog/fog/issues/3775

For now, I suggest you specify the dependency directly in your `Gemfile`

Issue #171, opened 2015-12-02
"uninitialized constant Fog::Storage::MIME"


celeen:
I am working on building a middleman static site to aws S3, and encountering the following error:
```gems/fog-core-1.35.0/lib/fog/storage.rb:68:in `get_content_type': uninitialized constant Fog::Storage::MIME (NameError)```

I've done a bit of tracing, and it looks to me like what's going on is:

mime/types is inaccessible in [lib/fog/storage.rb:68](https://github.com/fog/fog-core/blob/master/lib/fog/storage.rb#L68) at least when called from [fog-aws requests/storage/put_object.rb:43](https://github.com/fog/fog-aws/blob/master/lib/fog/aws/requests/storage/put_object.rb#L43) (where Fog::Storage#parse_data is called), presumably because [Fog::Storage#new](https://github.com/fog/fog-core/blob/master/lib/fog/storage.rb#L11) does not actually run in this instance.

Perhaps [the recent commit](https://github.com/fog/fog-core/commit/0fd7d8f846376fe3dc6d804163b0079a2b343226) that moved requirement of mime/types or mime/types/columnar inside self#new is what created this bug.

Would love feedback as to the accuracy of this diagnosis, and what preferable fix is. Happy to submit a PR with preferred fix if needed. Thanks!

plribeiro3000:
So there is 2 things here.

* You need to add the gem `mime-types` to your project
* Fog needs to correctly raise an exception telling you to do so

I'm not totally sure why it didn't. Maybe @geemus have a better idea?

geemus:
Hmm. @celeen I would think you would have to call new before calling put_object though? That would give the connection needed, though I could be wrong.

geemus:
Actually, upon further consideration I suspect that you have an older version of fog (which doesn't have the mime-type warning in Storage#new) paired with a new version of fog-core. Could you try updating fog and see if that helps?

celeen:
Oh, man, I've run bundle update a whole buuunnnnch of times in the past 12 hours. Maybe I'm not completely understanding how the fog architecture works together-- I'm using the gem middleman-s3_sync, which is calling on fog-aws (0.7.6), which is calling on fog-core(1.35.0). I've put some very sophisticated `p` statements into my local version of the gem fog-core to try to trace whether Storage.new gets called after running `middleman build` (which kicks off the s3_sync process), and no dice.

[...]

PR #221, opened 2017-11-13
Add missing runtime dependency declaration to gemspec

NARKOZ:
This adds mime-types as runtime dependency into gemspec, which is required in https://github.com/fog/fog-core/blob/master/lib/fog/storage.rb

It wasn't added before because of fog's support of Ruby 1.9.3. 
Support for Ruby 1.9.3 was removed from fog in https://github.com/fog/fog/pull/3974

### DIFF
--- a/fog-core.gemspec
+++ b/fog-core.gemspec
@@ -21,6 +21,7 @@ Gem::Specification.new do |spec|
   spec.required_ruby_version = '>= 2.0.0'
 
   spec.add_dependency("builder")
+  spec.add_dependency("mime-types")
   spec.add_dependency("excon", "~> 0.58")
   spec.add_dependency("formatador", "~> 0.2")
 
--- a/lib/fog/storage.rb
+++ b/lib/fog/storage.rb
@@ -1,20 +1,15 @@
+begin
+  # Use mime/types/columnar if available, for reduced memory usage
+  require 'mime/types/columnar'
+rescue LoadError
+  require 'mime/types'
+end
+
 module Fog
   module Storage
     extend Fog::ServicesMixin
 
     def self.new(orig_attributes)
-      begin
-        # Use mime/types/columnar if available, for reduced memory usage
-        require 'mime/types/columnar'
-      rescue LoadError
-        begin
-          require 'mime/types'
-        rescue LoadError
-          Fog::Logger.warning("'mime-types' missing, please install and try again.")
-          raise
-        end
-      end
-
       attributes = orig_attributes.dup # prevent delete from having side effects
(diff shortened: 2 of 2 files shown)


plribeiro3000:
Hey @NARKOZ , Thank you for this PR, but it believe we would need to remove all checks for `mime-types` in the code as well. Could you do that?

NARKOZ:
Done.

plribeiro3000:
Awesome! Thanks!

geemus:
I had also been somewhat reticent to force the dependency on everyone, when only storage users are likely to need it. It does simplify some things to require it, but at that cost I suppose. Does that make sense? What do you both think?

plribeiro3000:
I guess you are correct @geemus. Didn't think of it this way when i merged it.
One thing to mention is that probably `mime-types` is already a dependency at the project.

But i'm ok reverting this PR.

NARKOZ:
@geemus FYI, there have been many users who run into issues and complained about this:

https://github.com/fog/fog-core/issues/203

https://github.com/fog/fog-core/issues/177

https://github.com/fog/fog-core/issues/171

geemus:
@plribeiro3000 @NARKOZ thanks for the additional inputs, I'm fine keeping it in at this point (just wanted to discuss and make sure I wasn't missing anything). Thanks!

plribeiro3000:
@geemus although you said you are ok i tend to agree that removing the dependency makes sense specially when the final user is using an extracted provider.

I would love to have conditional dependency support out of the box at bundler/rubygems. 😄

geemus:
Yeah, I guess that is true. Maybe we should push the dependency out to the extracted providers instead of having it here? (though it will have the same effect in the short term). Even in the extracted case it might not be needed (like aws, which includes many things), but having a narrower inclusion might still be good. Is that what you were thinking?

plribeiro3000:
@geemus not what i was thinking but i like it. One thing to consider is that we might have breakeages if different providers pin different versions. Still, i like the idea to have dependencies like `mime-types` and `net-ssh` on providers that do use it instead of pushing it to everyone.

plribeiro3000:
Then, we can have checks (like this PR removed) just to tell the final user the provider (or at least the fog for that provider) does not support this feature.

geemus:
I'm fine just leaving this as-is for now. We can extract/push-out to the edges post-extraction if we still desire to do so. Alternatively we could create a fog-storage gem which had the requirement and do things that way. In any event, it solves a real problem for now, so being picky about how it does that seems like it might be a bit premature.

PR #223, opened 2018-02-20
Fix warning: key :libvirt_uri is duplicated

weppos:
```
/lib/fog/test_helpers/mock_helper.rb:69: warning: key :libvirt_uri is duplicated and overwritten on line 91
```

### DIFF
--- a/lib/fog/test_helpers/mock_helper.rb
+++ b/lib/fog/test_helpers/mock_helper.rb
@@ -66,7 +66,6 @@ if Fog.mock?
     :ovirt_url                        => "http://ovirt:8080/api",
     :ovirt_username                   => "admin@internal",
     :ovirt_password                   => "123123",
-    :libvirt_uri                      => "qemu://libvirt/system",
     :rackspace_api_key                => "rackspace_api_key",
     :rackspace_username               => "rackspace_username",
     :riakcs_access_key_id             => "riakcs_access_key_id",


plribeiro3000:
Thanks @weppos !

geemus:
Thanks!

weppos:
@plribeiro3000 @geemus you're welcome.

I actually have one question: I see in the mock helper there are several configs, some referencing libraries that have been extracted, such as `fog-dnsimple`:

https://github.com/fog/fog-core/blob/30b488b99f47072eba32d2a1e656d538cd80f4c8/lib/fog/test_helpers/mock_helper.rb#L30-L31

Is it safe to remove them?

plribeiro3000:
I'm might be wrong but i think they are needed for test purposes on the extracted providers, unless they can infer this data themselves. Didn't try so can say it works. =s

geemus:
Yeah, I think @plribeiro3000 might be right. Rather than removing, we need to migrate so that the extracted things can re-add things (I think). Which doesn't have a clear pattern yet, I don't think. In the mean time, it's not great but (hopefully) doesn't really hurt anything for it to be here.

weppos:
OK. I'm asking because the DNSimple ones are outdated and to be fair they are not even required in the real gem specs. I will provide the updated ones.

PR #224, opened 2018-02-20
Update DNSimple auth variables

weppos:
See: https://github.com/fog/fog-dnsimple/blob/be61ab5b089263893879ad36b944cb25ef88f6dc/lib/fog/dnsimple/dns.rb#L55-L56


plribeiro3000:
Awesome. Thanks!

geemus:
Thanks!

PR #225, opened 2018-03-02
Introduce kubevirt provider

pkliczewski:
We update provider dependencies to enable [kubevirt](http://www.kubevirt.io/) provider.


geemus:
This one is just doc changes basically, so going to go ahead and bring it in in prep for the other stuff. Thanks!

PR #228, opened 2018-03-07
fix nil fetch on object reload

lanej:
There is no guarantee that `collection.get` will return an object and the `nil` case should be handled appropriately.

### DIFF
--- a/lib/fog/core/model.rb
+++ b/lib/fog/core/model.rb
@@ -46,6 +46,8 @@ module Fog
 
       object = collection.get(identity)
 
+      return unless object
+
       merge_attributes(object.all_associations_and_attributes)
 
       self


lanej:
References https://github.com/fog/fog-aws/pull/433

plribeiro3000:
Gotcha. Makes sense. My only question is if we shouldn't raise an exception of at least throw a log?

WDYT?

plribeiro3000:
Hmmm. I just realized the current approach already returns nil if an `excon` error happens.

Maybe we can leave as is or just revisit the whole method?

geemus:
Get is expected to return nil if nothing is found, which ends up happening with aws due to eventual consistency (if you try to look it up right after creating it, you often get not founds for a few seconds). I think if reload returns falsey, we should be in an ok place. I do wonder if it warrants a warning or something, but at least in the aws usage it would usually be hidden inside a wait_for (which will inherently smooth off the missing messages). I'm fine bringing this in as is for now.

@lanej thanks for taking the time to track that down.

lanej:
Thanks to all for wrapping this up quickly.

geemus:
released as part of v2.1.0, thanks!

PR #229, opened 2018-04-02
Force Gem Best Practice

plribeiro3000:
All classes defined from within a Gem should be inside its own namespace
to avoid conflicts with other gems

cc/ @geemus @icco @Temikus @lanej @fog/core 

This will be the first PR of many to
* Achieve minimum consistency across all providers
* Implement some best practices
* Improve loading of the libraries by using bundle instead of manually requiring it

### DIFF
--- a/lib/fog/core/provider.rb
+++ b/lib/fog/core/provider.rb
@@ -21,11 +21,21 @@ module Fog
       Fog.services[new_service] ||= []
       Fog.services[new_service] |= [to_s.split("::").last.downcase.to_sym]
       @services_registry ||= {}
-      @services_registry[new_service] = [to_s, constant_string].join("::")
+      @services_registry[new_service] = service_klass(constant_string)
     end
 
     def services
       @services_registry.keys
+    end
+
+    def service_klass(constant_string)
+      eval([to_s, constant_string].join("::"))
+      [to_s, constant_string].join("::")
+    rescue NameError
+      provider = to_s.split("::").last
+      Fog::Logger.deprecation("Unable to load #{[to_s, constant_string].join("::")}")
+      Fog::Logger.deprecation("The format #{['Fog', constant_string, provider].join("::")} is deprecated")
+      ['Fog', constant_string, provider].join("::")
     end
   end
 end
--- a/lib/fog/core/services_mixin.rb
+++ b/lib/fog/core/services_mixin.rb
@@ -24,16 +24,24 @@ module Fog
 
     private
 
+    # This method should be removed once all providers are extracted.
+    # Bundler will correctly require all dependencies automatically and thus
+    # fog-core wont need to know any specifics providers. Each provider will
+    # have to load its dependencies.
     def require_service_provider_library(service, provider)
       require "fog/#{provider}/#{service}"
     rescue LoadError  # Try to require the service provider in an alternate location
+      Fog::Logger.deprecation("Unable to require fog/#{provider}/#{service}")
+      Fog::Logger.deprecation("The format fog/#{service}/#{provider} is deprecated")
       require "fog/#{service}/#{provider}"
     end
 
     def service_provider_constant(service_name, provider_name)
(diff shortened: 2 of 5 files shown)


lib/fog/core/provider.rb
```diff
@@ -21,11 +21,20 @@ def service(new_service, constant_string)
       Fog.services[new_service] ||= []
       Fog.services[new_service] |= [to_s.split("::").last.downcase.to_sym]
       @services_registry ||= {}
-      @services_registry[new_service] = [to_s, constant_string].join("::")
+      @services_registry[new_service] = service_klass(constant_string)
```
plribeiro3000: I'm not sure how this is working now. As of my understanding it will always load services in the form of `Fog::<PROVIDER>::<Service>` and we have many providers doing `Fog::<Service>::<PROVIDER>`

geemus: If I recall correctly, it was all `Fog::<PROVIDER>::<SERVICE>` initially, and we might have changed some along the way in the hopes that the other way was better? It would be best, I suspect, to pick one and stick to it though (which would probably be the old way). What do you think?

lib/fog/core/services_mixin.rb
```diff
@@ -24,16 +24,22 @@ def providers
...
     private
 
+    # This method should be removed once all providers are extracted.
+    # Bundler will correctly require all dependencies automatically and thus
+    # fog-core wont need to know any specifics providers. Each provider will
+    # have to load its dependencies.
     def require_service_provider_library(service, provider)
       require "fog/#{provider}/#{service}"
     rescue LoadError  # Try to require the service provider in an alternate location
+      Fog::Logger.deprecation('Prefer to define all Services within the Provider Namespace instead!')
```
geemus: I think maybe a more specific error would be helpful here. Like `Unable to load "#{fog/#{provider}/#{service}"`. What do you think?

lib/fog/core/services_mixin.rb
```diff
@@ -24,16 +24,22 @@ def providers
...
       require "fog/#{service}/#{provider}"
     end
 
     def service_provider_constant(service_name, provider_name)
-      Fog.const_get(service_name).const_get(*const_get_args(provider_name))
-    rescue NameError  # Try to find the constant from in an alternate location
       Fog.const_get(provider_name).const_get(*const_get_args(service_name))
+    rescue NameError  # Try to find the constant from in an alternate location
+      Fog::Logger.deprecation('Prefer to define all Services within the Provider Namespace instead!')
```
geemus: Similarly, a more specific error here would probably be less confusing, though I'm much less sure how to word it helpfully here. What do you think?

spec/spec_helper.rb
```diff
@@ -10,6 +10,7 @@
 require "minitest/autorun"
 require "minitest/spec"
 require "minitest/stub_const"
+require "pry"
```
geemus: I suspect it may have been unintentional to commit this?

lib/fog/core/provider.rb
plribeiro3000: I agree to sticking to one. So my idea is to deprecate the one that might give us problems (2 different providers define a class named `Fog::Compute::Utils`) and drop support in a major version. This is being done from the `service_klass` method.

lib/fog/core/services_mixin.rb
plribeiro3000: Hmmm. Good idea. How to announce the deprecation as well?

spec/spec_helper.rb
plribeiro3000: Ops. Good catch. =)

plribeiro3000:
Will work on your comments later today. Thanks for the reviewW

plribeiro3000:
Here we go @geemus. I ended up doing 2 messages, one to tell it could not load the class and another one to notify about the deprecation.

What do you think about this approach? Maybe too verbose?

geemus:
Looks good, thanks!

PR #238, opened 2018-08-13
Fix `@path_prefix` not initialized warning

tokengeek:
Setting `@path_prefix` within a condition meant it was not initialized
to `nil` in cases when it was not passed in. This triggered warnings
when Ruby was showing them and the variable was used.

`Hash#delete` already returns `nil` when the key is missing so that can
be used to initialize it correctly.


plribeiro3000:
Good catch!

geemus:
Thanks!

PR #243, opened 2018-09-09
Do not force number option values as numbers - fix for OVH OpenStack

cyrylas:
This fixes OVH openstack support.

In OVH tennant / project name is a number. When option is converted into number we get error:
```
{"error": {"message": "object of type 'int' has no len()", "code": 400, "title": "Bad Request"}}
```
I think, that fog should not enforce variable types. When user specifies something as string - it should be handled string.

### DIFF
--- a/lib/fog/core/service.rb
+++ b/lib/fog/core/service.rb
@@ -163,8 +163,6 @@ module Fog
           value_string = value.to_s.downcase
           if value.nil?
             options.delete(key)
-          elsif value_string.to_i.to_s == value
-            options[key] = value.to_i
           else
             options[key] = case value_string
                            when "false"
--- a/spec/service_spec.rb
+++ b/spec/service_spec.rb
@@ -58,10 +58,16 @@ describe Fog::Service do
       refute_includes service.options.keys, :generic_user
     end
 
-    it "converts number String values with to_i" do
+    it "handles number String as numbers" do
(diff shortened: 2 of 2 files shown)


tokengeek:
The change that is made is in `coerce_options` which from its name is intended to take options and convert them into Ruby types, including numbers. So this is established behaviour relied on by all providers. So this change will probably break more things than help, as demonstrated by needing to change the tests.

It actually sounds like a fix is required in https://github.com/fog/fog-openstack. Either `coerce_options` is used where is shouldn't be or the tenant needs processing.

It's `fog-openstack`'s responsibly to make sure that it is making API requests with the correct JSON so the server does not attempt to call `len` on a number. Assuming that error response is from OpenStack then it seems like `to_s` is needed in some places because it is generating bad requests.

Is there an issue reported for this in fog-openstack yet?

geemus:
@cyrylas I would tend to agree with @tokengeek that this would be difficult and dangerous to change here. Are there particular calls in fog-openstack where this occurs? I agree that this is happening is unfortunate, but I think making a more localized change to to_s the strings on the openstack side seems safer. Does that make sense?

cyrylas:
Yes, you're right. I though this is some new change, but indeed it old one. I'll try to fix that in fog-openstack.

geemus:
@cyrylas No worries, thanks for understanding. Just let us know if you need further assistance.

PR #249, opened 2019-02-23
Log issues when loading providers

iNecas:
I was hitting issue Fog::Service::NotFound: ovirt has no compute service
when trying to use fog-ovirt and there was no indication of what could be
wrong. Eventually, I've figured out, that there was issue with
dependencies inside fog-ovirt, that we were suppressing.

This patch adds logging around the requiring the provider so that this
issues are easier to debug later.

There are several issues in the past around the 'no compute service' that would benefit from more logging,
such as:

- https://github.com/fog/fog-xenserver/issues/60
- https://github.com/fog/fog-google/issues/170
- https://github.com/fog/fog-openstack/issues/181
- https://github.com/mitchellh/vagrant-aws/issues/505

### DIFF
--- a/lib/fog/core/services_mixin.rb
+++ b/lib/fog/core/services_mixin.rb
@@ -27,7 +27,9 @@ module Fog
       require_service_provider_library(service_name.downcase, provider)
       spc = service_provider_constant(service_name, provider_name)
       spc.new(attributes)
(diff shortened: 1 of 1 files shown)


plribeiro3000:
Awesome. Thanks!

geemus:
Thanks!

PR #253, opened 2020-01-16
Use :never in ssh/scp verify_host_key option

### LINKED ISSUE #251: verify_host_key: false is deprecated, use :never
Hi, I see this error always I use Fog::SSH...maybe it could be convenient to change the code of ssh.rb to use the :never mentioned in the warning, don't it?

militarpancho:
Fix #251

### DIFF
--- a/lib/fog/core/scp.rb
+++ b/lib/fog/core/scp.rb
@@ -64,8 +64,9 @@ module Fog
 
         @address  = address
         @username = username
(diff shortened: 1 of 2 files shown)


geemus:
This looks perfect, thanks for taking the time to do this (and adding the extra cautious backwards compatible bits, just in case).

militarpancho:
it was a pleasure!!

PR #255, opened 2020-03-04
Auto coerce name key in config into string

nicolasleger:
Currently if the name of the project is like a number like `123` for example,  passing `"123"` as `openstack_project_name` option will not return an error from the remote OpenStack server.
Because it is auto coerced into `integer` and not `string`.

This patch forces `openstack_project_name` and alike params to be string.

### DIFF
--- a/lib/fog/core/service.rb
+++ b/lib/fog/core/service.rb
@@ -163,6 +163,8 @@ module Fog
           value_string = value.to_s.downcase
           if value.nil?
             options.delete(key)
+         elsif key.end_with? 'name'
+            options[key] = value.to_s
           elsif value_string.to_i.to_s == value
             options[key] = value.to_i
           else


lib/fog/core/service.rb
```diff
@@ -163,6 +163,8 @@ def coerce_options(options)
           value_string = value.to_s.downcase
           if value.nil?
             options.delete(key)
+         elsif key.end_with? 'name'
+            options[key] = value.to_s
```
plribeiro3000: I'm afraid this behaviour might break other providers that rely on this, no?

lib/fog/core/service.rb
```diff
@@ -163,6 +163,8 @@ def coerce_options(options)
           value_string = value.to_s.downcase
           if value.nil?
             options.delete(key)
+         elsif key.end_with? 'name'
```
plribeiro3000: Ins't `end_with` available only on ruby `2.6+`?

nicolasleger: Seems not (checked ✅on 2.3) https://apidock.com/ruby/String/end_with%3f

lib/fog/core/service.rb
nicolasleger: One provider I use can have a "123" project name currently, and will reject the `123` integer despite filling `"123"` string as option.

I might restrict the check to `elsif value.is_a?(String) && key.end_with? 'name` to be sure, what do you think?

plribeiro3000: Since this is an openstack issue i rather make this change for it only since we have several dozens. WDYT?

geemus: Yeah, this has the potential for a much broader impact than we really desire. I would suggest we try to make a change in open-stack itself, where we could explicitly check for integer values in project name and raise a specific error. Does that make sense?

PR #258, opened 2020-09-15
Fix Call to `#underscore_name`

tubbo:
This method was defined on the instance level, not the class level, so `Provider.extends` couldn't get to it. Broke our build when we attempted to update to v2.2.1.

### DIFF
--- a/lib/fog/core/provider.rb
+++ b/lib/fog/core/provider.rb
@@ -8,10 +8,22 @@ module Fog
   end
 
   module Provider
-    def self.extended(base)
-      provider = base.to_s.split("::").last
-      Fog.providers[provider.downcase.to_sym] = provider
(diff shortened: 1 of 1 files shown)


geemus (approved):
Thanks for the quick fix!

geemus:
Released in 2.2.2

tubbo:
@geemus noticed that you folks have tests but they don't run when PRs are opened. you might want to fix that? last PR would have failed tests if this was the case.

geemus:
Yeah, I noticed that. I think all the other fog stuff has CI setup, but we haven't touched this part in some time (prior to today the last release was more than a year ago), so I think maybe it slipped through the cracks. Will work to remedy.

geemus:
Added basic ruby test runner to actions, so hopefully we should be good to go on that.

PR #259, opened 2020-09-15
fix(Provider): lookup

plribeiro3000:
@geemus Now that we got that release i was able to figure out what was wrong with the loading at fog/fog-internet-archive#2.

~~I will point that PR to this branch and see how it goes. Will check here later.~~

EDIT: 

It works now. we had two minor issues:

* The alias provider name was not a symbol
* The guard clause was not looking for the non aliased name

### DIFF
--- a/lib/fog/core/provider.rb
+++ b/lib/fog/core/provider.rb
@@ -12,7 +12,7 @@ module Fog
       def extended(base)
         provider = base.to_s.split("::").last
         Fog.providers[provider.downcase.to_sym] = provider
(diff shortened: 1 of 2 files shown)


geemus:
@plribeiro3000 Thanks! Is this ready to go then? And will you need another release?

plribeiro3000:
@geemus yeap to both questions. =D

geemus:
Done (2.2.3)

plribeiro3000:
Awesome. Thanks @geemus !

PR #261, opened 2021-01-20
PERF: Eliminate dependency on mime-types gem

BigBigDoudou:
**Inspired by: https://github.com/rest-client/rest-client/pull/644**

Eliminate the dependency on `mime-types` which is a memory hog.

The gem `mini_mime` (https://github.com/discourse/mini_mime) was created in replacement. It uses the exact same database as the full fledged mime types and is capable of only loading stuff on demand with a practical, safe and bound in-memory cache.

Allocated memory by gem (derailed benchmarks):

| gem | memory |
|---|---|
| mime-types | 6258796 | 
| mini_mime | 7606 |

### DIFF
--- a/fog-core.gemspec
+++ b/fog-core.gemspec
@@ -21,7 +21,7 @@ Gem::Specification.new do |spec|
   spec.required_ruby_version = '>= 2.0.0'
 
   spec.add_dependency("builder")
-  spec.add_dependency("mime-types")
+  spec.add_dependency('mini_mime', '>= 0.1.1')
   spec.add_dependency("excon", "~> 0.71")
   spec.add_dependency("formatador", "~> 0.2")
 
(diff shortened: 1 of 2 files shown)


BigBigDoudou:
The CI seems to be broken due to depreciation of set-env and add-path commands :/ (https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/)

plribeiro3000:
I wonder if this change might break a provider that has `mimi-types` hardcoded. 🤔 

Will try to check it.

BigBigDoudou:
Hi @plribeiro3000 , did you find anything finally?

geemus:
@BigBigDoudou Thanks for this. I share @plribeiro3000 concern about it causing other issues without some other changes though. Looking around at the other fog repos I see a similar require for the columnar mime-types stuff in a few places (which would presumably now fail), so the transition is probably worth doing, but also seems likely to be a bit more involved than this. Does that make sense? Would you be up for helping with the outlying changes also? The challenge then would become one of the timing/coordination for the releases, but we can cross that bridge when we reach it.

PR #269, opened 2021-07-31
Avoid loading unnecessary service via autoload

pocke:
This change avoids loading unnecessary service classes when the service is defined with `autoload`, such as fog-aws.

# Problem

fog-aws  uses `autoload` to load each service to avoid loading unnecessary service. By the `autoload`s, the end-users can reduce startup time and memory usage.
But the `autoload`s don't work actually. fog-aws loads all services even if the user doesn't need all services.

Because fog-core evaluates all services' constants with `eval` method. The evaluation loads unnecessary code unexpectedly.

# Solution

Use `const_efined?` instead of `eval`. `const_defined?` method checks constant existence without loading the code.

Note that I also reorder class definitions in the test. It reduces unexpected warnings, such as `Unable to load ...`. It doesn't the test result, so I'll revert the test changes if you don't need it.

# Benchmark

I benchmarked memory usage and startup time with `fog-aws` package.

## memory usage

It reduces about 585KB, 9%.

### before

```bash
$ RUBYLIB=lib ruby -robjspace -e 'require "fog/aws";GC.start; puts "#{ObjectSpace.memsize_of_all * 0.001 * 0.001} MB"'
6.844114 MB
```

### after

```bash
$ RUBYLIB=lib ruby -robjspace -e 'require "fog/aws";GC.start; puts "#{ObjectSpace.memsize_of_all * 0.001 * 0.001} MB"'
6.258867 MB
```

## Startup time

It reduces about 45ms, 9%.

### before

```bash
$ RUBYLIB=lib ruby -robjspace -rbenchmark -e 'p(Benchmark.realtime {require "fog/aws"})'
0.5202090488746762
```

### after

```bash
$ RUBYLIB=lib ruby -robjspace -rbenchmark -e 'p(Benchmark.realtime {require "fog/aws"})'
0.47530297795310616
```

### DIFF
--- a/spec/compute_spec.rb
+++ b/spec/compute_spec.rb
@@ -10,29 +10,22 @@ describe "Fog::Compute" do
   describe "#new" do
     module Fog
       module TheRightWay
-        extend Provider
-        service(:compute, "Compute")
-      end
-    end
-
-    module Fog
-      module TheRightWay
         class Compute
           def initialize(_args); end
         end
       end
     end
 
-    it "instantiates an instance of Fog::Compute::<Provider> from the :provider keyword arg" do
-      compute = Fog::Compute.new(:provider => :therightway)
-      assert_instance_of(Fog::TheRightWay::Compute, compute)
-    end
-
     module Fog
-      module TheWrongWay
+      module TheRightWay
         extend Provider
         service(:compute, "Compute")
       end
+    end
+
+    it "instantiates an instance of Fog::Compute::<Provider> from the :provider keyword arg" do
+      compute = Fog::Compute.new(:provider => :therightway)
+      assert_instance_of(Fog::TheRightWay::Compute, compute)
     end
 
     module Fog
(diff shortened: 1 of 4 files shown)


spec/compute_spec.rb
```diff
@@ -10,16 +10,16 @@ def self.require(*_args); end
...
       end
     end
 
     module Fog
       module TheRightWay
-        class Compute
-          def initialize(_args); end
-        end
+        extend Provider
+        service(:compute, "Compute")
```
plribeiro3000, line 22: I don't get why you've changed the order here.
Is it because of some test failure?

spec/compute_spec.rb
```diff
@@ -74,6 +67,13 @@ def initialize(_args); end
       end
     end
 
+    module Fog
+      module BothWays
+        extend Provider
+        service(:compute, "Compute")
+      end
+    end
```
plribeiro3000, line 75: Same thing here. The code you've changed should not break this tests, should it?

plribeiro3000:
@pocke Thats a great improvement you did here, specially with the extra benchmarks but i'm confused as why did you move some pieces of code around in the tests.

pocke:
Thanks for your review. 


> but i'm confused as why did you move some pieces of code around in the tests.


Sorry for my lack of describes. The order of the classes doesn't affect the test result, it is just a refactoring. Previously the service classes are defined before `service(:name, 'ClassName')` is called. It introduces deprecation warnings such as `[fog][DEPRECATION] Unable to load Fog::TheRightWay::Compute`.

I think the deprecation warnings are not intentional, and I'd like to confirm the patch works on the appropriate situation, so I changed the order.

geemus (approved):
Looks good. Thanks for the detailed report and fix.

PR #271, opened 2021-09-14
Add Truffleruby head to CI

gogainda:
Hi, I notice that MRI Ruby head is failing here, but Truffleruby head passes

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -19,7 +19,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


geemus:
@gogainda Thanks, I've added head to allowed failures to make sure the rest runs, at least for now. It appears truffleruby is failing when it tries to bundle install gems though. Could you take another look?

gogainda:
I dont see the failure, were do you see it?

geemus:
Never mind, I think I was misreading it. Sorry about that. It is still being cancelled due to the head failure, as your branch doesn't have the ignore yet. I'll go ahead and pull this in for now. Thanks!

PR #280, opened 2023-04-19
Cache and restore collection attributes

jsirex:
Some collections also have attributes. It is better to cache collection's attributes too and properly restore model and collection.

For example, DNS `Zone` model has `Records` collection. This collection requires `zone_id` attribute to operate: get all records by `zone_id`, get record by `zone_id` and record `identity`, etc. When we load from cache a `Record` Fog creates collection for this model without any attributes.

Also when cache is not `valid_for_load?` let display which keys are missing.

### DIFF
--- a/lib/fog/core/cache.rb
+++ b/lib/fog/core/cache.rb
@@ -93,6 +93,9 @@ module Fog
     # see the +namespace_prefix=+ method.
     SANDBOX = ENV["HOME"] ? File.expand_path("~/.fog-cache") : File.expand_path(".fog-cache")
 
+    # required attributes to load model from cache
+    REQUIRED_ATTRIBUTES = [:identity, :collection_klass, :collection_attrs, :model_klass, :model_attrs]
+
     # when a resource is used such as `server.cache.dump` the model klass is passed in
     # so that it can be identified from a different session.
     attr_reader :model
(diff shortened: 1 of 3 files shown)


jsirex:
Failed tests are not related to the change.

``` diff
-                raise_error = lambda { |_cmd, _options| raise Net::SSH::AuthenticationFailed.new }
+                raise_error = lambda { |_cmd, _options={}| raise Net::SSH::AuthenticationFailed.new }
```

geemus:
@jsirex - I think this looks good overall. Could you add some associated tests to the cache tests as well?

jsirex:
Done. 

But I think we need further improvements of the Cache class.
For example, tests wipe out .fog-cache folder and littering home directory.
Also, I believe SANDBOX must not be constant.
Probably I'll make a new PR

geemus:
Yeah, those are good callouts. We should probably make it more configurable and in tests maybe use a tmp directory or something to avoid the problems you mentioned. I guess in tests we could set `ENV['HOME']` to a tmpdir already to get this? Agreed that how this is setup now is a bit strange, I suspect this code may not get used the most by others given some of the quirks you have pointed out.

geemus (approved):

geemus:
Thanks!

PR #281, opened 2023-05-12
Model methods

jsirex:
Please, do not squash commits when (if) merging.

Add methods `#create`, `#update`, `#save`, `#destroy` explicitly.
Be more consistent with return values.
Add `filter_attributes` helper method.

Commit messages below:

---
Model methods like #save or #destroy are implicit: they only mentioned
in README but explicitly used by framework. Let define them also
explicitly so developer knows exactly what to implement.

This change is not breaking change, because either method is already
defined in child class or is not used at all.

Another goal is to make return values consistent.

---
It is just handy method to easily get a sub-set of attributes.
For example, my model has 10+ attributes, but when creating/updating
model I need only a few of them:

Code:
    ...
    attribute :name
    attribute :value
    attribute :comment
    attribute :type
    ...
    service.create_entity(filter_attributes(:name, :value, :type))

### DIFF
--- a/lib/fog/core/model.rb
+++ b/lib/fog/core/model.rb
@@ -21,6 +21,33 @@ module Fog
       merge_attributes(attribs)
     end
 
+    # Creates new or updates existing model
+    # @return [self]
+    def save
+      persisted? ? update : create
+    end
+
+    # Creates new entity from model
+    # @raise [Fog::Errors::NotImplemented] you must implement #create method in child class and return self
+    # @return [self]
+    def create
+      raise Fog::Errors::NotImplemented, "Implement method #create for #{self.class}. Method must return self"
+    end
+
+    # Updates new entity with model
+    # @raise [Fog::Errors::NotImplemented] you must implement #update method in child class and return self
+    # @return [self]
+    def update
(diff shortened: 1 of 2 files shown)


jsirex:
Also I have some questions:
1. Should we drop any code related to ruby 1.x? According to README 2.x now required
2. Should I update rubocop configuration to use 2.x syntax?

geemus:
Thanks for the updates. With `filter_attributes` it's a method that you would call on a instance, right? So it might look something like this:
```
existing_object = service.collection.all.first
copycat = service.collection.create(existing_object.filter_attributes(:name, :value))
```
I think that makes sense to me, I wasn't exactly following the initial example above, so I wanted to check in. Is something like what I wrote what you had in mind, or am I still missing something?

I think it would be fine to drop 1.x code and update rubocop, but it may make sense to do that in distinct PRs.

jsirex:
Let it be wip. I'll split out logic in other PRs. Just busy a bit last couple of days.

geemus:
Yeah, no worries and no hurry, just wanted to see what you had in mind.

Just let me know when you get a chance and I'll review.

jsirex:
I think this PR is ready for review/merge

geemus (approved):
LGTM, thanks!

PR #282, opened 2023-05-22
Add filter_attributes method

jsirex:
It is just handy method to easily get a sub-set of attributes. For example, my model has 10+ attributes, but when creating/updating model I need only a few of them:

Code:
    ...
    attribute :name
    attribute :value
    attribute :comment
    attribute :type
    ...
    service.create_entity(filter_attributes(:name, :value, :type))

### DIFF
--- a/lib/fog/core/attributes.rb
+++ b/lib/fog/core/attributes.rb
@@ -25,6 +25,14 @@ module Fog
         @masks ||= {}
       end
 
+      # Defines a new attribute
+      # @param [Symbol] name attribute name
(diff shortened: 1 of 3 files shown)


jsirex:
ps. if you merge #281 I then rebase this

plribeiro3000:
Nice work on this @jsirex .

I'm happy to see someone else upgrading the part of the core of `fog`. ❤️ 💚 💙 💜 💛 🖤 🌈

geemus (approved):
LGTM, thanks!

PR #283, opened 2023-05-22
Update rubocop configuration

jsirex:
Let start with discussion about rubocop configuration.

I think it is suitable for us:
``` yaml
Style/HashSyntax
  EnforcedStyle: ruby19_no_mixed_keys
```

Double quoutes appear everywhere:

``` yaml
Style/StringLiterals:
  EnforcedStyle: double_quotes
```

I personally prefer single quotes like this:
``` yaml
Style/StringLiterals:
  Enabled: true
  EnforcedStyle: single_quotes

Style/StringLiteralsInInterpolation:
  Enabled: true
  EnforcedStyle: single_quotes
```
but it need to be discussed.


tokengeek:
The quotes style was discussed when we first added Rubocop to https://github.com/fog/fog/

In person, when half a dozen of us met up at a fog summit to discuss the project direction.

Since it is a personal preference, we decided on one and set it up to avoid multiple styles across many repos when we started breaking them apart.

I can see no value changing style now for a personal preference, especially since it would require a noisy commit to swap everything from double to single quotes.

jsirex:
Ok, @tokengeek 
What should be for Style/StringLiteralsInInterpolation?

By default it "EnforcedStyle: single_quotes (default)". I want to explicitly add it to the rubocop configuration.

fog-core.gemspec
```diff
@@ -15,26 +16,29 @@ Gem::Specification.new do |spec|
...
   spec.add_development_dependency("pry")
   spec.add_development_dependency("rake")
   spec.add_development_dependency("rubocop")
+  spec.add_development_dependency("rubocop-minitest")
+  spec.add_development_dependency("rubocop-rake")
   spec.add_development_dependency("thor")
+  spec.add_development_dependency("tins")
   spec.add_development_dependency("yard")
+
+  spec.metadata["rubygems_mfa_required"] = "true"
```
jsirex, line 43: @geemus, rubocop insists on rubygems mfa. Is it ok? Do you have mfa enabled?

geemus:
I don't recall exactly what we discussed on quoting, but I'd prefer for us to not break a bunch of stuff a this point. That being said, it does seem like it would be nice to be explicit, instead of just falling back to an implicit rule.

jsirex:
If it looks good for you we can merge this commit.
There is rubocop_todo, so we can iteratively fix rubocop's offences in separate commits.

README.md
```diff
@@ -6,7 +6,7 @@ Shared classes and tests for fog providers and services.
 
 ## Ruby version
 
-Fog-core requires Ruby `2.0.0` or later.
+`fog-core` requires Ruby `2.5.0` or later.
```
plribeiro3000, line 9: Wait, this is a huge change.

I recall that we had to wait a long time before dropping ruby `1.9` because of LTS distros that still had support for it.

Is there a reason (besides the obvious non maintained ruby version) for this change?

plribeiro3000: Maybe @geemus can chime in since he was the one who told me to wait back then?

jsirex: I checked https://www.ruby-lang.org/en/downloads/branches/
```
status: eol
release date: 2019-12-25
normal maintenance until: 2022-04-01
EOL: 2023-03-31
Ruby 2.6

status: eol
release date: 2018-12-25
normal maintenance until: 2021-04-01
EOL: 2022-04-12
Ruby 2.5

status: eol
release date: 2017-12-25
normal maintenance until: 2020-04-01
EOL: 2021-04-05
```

I thought that older versions of ruby should not be used. Maybe we should stick to >= 2.7.
Actually, this all about rubocop rules + compatibility code.

Why I chose 2.5? Because fog-core has tests starting from 2.5.
But, if you think, that we *must* stick to 2.0, I can revert this part of changes.

Do you think we still have <2.5 users? 
Should next release in 2k23 (2k24) still support <2.5 ruby?

plribeiro3000: Last time i checked we had a lot of users using old distros in production that had LTS support for older versions of ruby but i dunno how it is nowadays.

If i remember correctly it was CentOS LTS and RedHat LTS that hold us true to ruby 2.0.

plribeiro3000: I guess that since our need to update the version of ruby is just because of rubocop, maybe we could freeze to a specific version of rubocop instead?

At leat for now?

And later we can tackle this on a different PR, with focus on updating ruby version?

jsirex: Ok, LTS distros with 2.0 is a good point. I'll revert this. Anyway, we must keep in mind, that tests actually running under 2.5 and higher.

geemus:
@plribeiro3000 other than the ruby version, any other thoughts?

I think for many of them we can probably work through them and figure out whether we want to make the changes or change the rules, rather than having to necessarily figure it all out up front, but I'm open to other ideas and approaches.

jsirex:
@geemus , I think we can start from this:
1. There are sane defaults in rubocop
2. It still configured to 2.0
3. Merge this change
4. there are 100+ offenses to be corrected 
5. I'll correct them one by one in future commits
6. Optionally: integrate rubocop check into the pipeline, so new commits can be validated too
7. If, eventually, we decide to change rubocop setting, then repeat this list: discuss, update config, regenerate todo, fix.
8.

geemus:
@jsirex - Yeah, that sounds good to me. Thanks for taking this on.

plribeiro3000:
LGTM @geemus @jsirex

PR #286, opened 2023-06-16
Fix wrong Cache#namespace path

jsirex:
SANDBOX and namespace_prefix can have their own register. We shouldn't downcase whole path.

Without this patch
Case 1: $HOME == '/home/Domain_User' - will fail (mkdir - permission denied /home/domain_user)
Case 2: namespace_prefix == 'MyService' - will create ~/.fog-cache/myservice


geemus (approved):

geemus:
Thanks!

PR #287, opened 2023-06-22
rubocop: Fix Layout/HashAlignment

jsirex:
Next step, but needs to be discussed:
I see in the project _table_ style, so decided that this style is preferred for Hashes.

Pay attention, that defaults (and, probably preferred by community) are:
EnforcedHashRocketStyle: key (default)
EnforcedColonStyle: key (default)

Merge or ask for change :)


jsirex:
Updated MR, added more commits.

54f82a2 * rubocop: Fix typo in method options - I believe fix is correct.
0454225 * rubocop: Lint/UselessAccessModifier - pay attention on this commit. "private" was removed.

jsirex:
@geemus we can stick to the defaults, almost zero effort to change this. Just let me know

geemus:
@jsirex In that case, maybe let's go with the defaults? As long as this only applies to fog-core I think that should be good. (applying across the suite of fog libraries could get to be a bit much).

jsirex:
Should I update anything else? Or it is ok to merge?

geemus (approved):
LGTM, thanks!

PR #289, opened 2023-08-04
Fix compatibility with Minitest 5.19+

voxik:
The `MiniTest` was renamed to `Minitest`:

https://github.com/minitest/minitest/commit/9a57c520ceac76abfe6105866f8548a94eb357b6

And the `MiniTest` constant is now loaded just when `MT_COMPAT` environment variable is set:

https://github.com/minitest/minitest/commit/a2c6c18570f6f0a1bf6af70fe3b6d9599a13fdd6

### DIFF
--- a/lib/fog/test_helpers/minitest/assertions.rb
+++ b/lib/fog/test_helpers/minitest/assertions.rb
@@ -1,6 +1,6 @@
 require "fog/schema/data_validator"
 
-module MiniTest::Assertions
+module Minitest::Assertions
   # Compares a hash's structure against a reference schema hash and returns true
   # when they match. Fog::Schema::Datavalidator is used for the validation.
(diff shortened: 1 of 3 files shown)


voxik:
Just FTR, this fixes test failures such as:

~~~
  1) Error:
Fog::Service::when config object can configure the service itself#test_0001_ignores the global and its values:
NameError: uninitialized constant MiniTest
    /builddir/build/BUILD/fog-core-2.2.4/usr/share/gems/gems/fog-core-2.2.4/spec/service_spec.rb:175:in `block (3 levels) in <top (required)>'
~~~

And I assume that the fixes in test helpers might help to other dependent projects (while I don't know where those are used in practice).

geemus (approved):
Thanks, yeah I saw this change on another project, but hadn't gotten to this yet.

PR #292, opened 2023-12-22
Delegate methods that accept keywords from Fog::Collection to Array c…

DimitriosLisenko:
…orrectly for Ruby 2.x and Ruby 3.x

### DIFF
--- a/lib/fog/core/collection.rb
+++ b/lib/fog/core/collection.rb
@@ -19,6 +19,7 @@ module Fog
           end
           super
         end
+        ruby2_keywords(method) if respond_to?(:ruby2_keywords, true)
       EOS
     end
 
@@ -31,6 +32,7 @@ module Fog
           data = super
           self.clone.clear.concat(data)
         end
+        ruby2_keywords(method) if respond_to?(:ruby2_keywords, true)
       EOS
     end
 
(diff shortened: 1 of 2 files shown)


geemus:
Oops, I was trying to do too many things at once and merged this before I was ready.

I just created #294 to bump the test matrix so that CI will use 3.2 (this is what I had intended to merge and just didn't pay enough attention to which browser tab I was in). I guess I had expected from reading the issue information that we could have expected failures in 3.2 that this change could then fix. Are there tests we could add that would fail without the test and pass with them? I think that would give me a bit more confidence that we have fixed this (and don't break it again in the future). Does that make sense? Thanks.

DimitriosLisenko:
Hey @geemus! No worries, it happens :) I'm afraid I don't follow re: the tests, could you please clarify what kind of tests you'd like added? The test I have already added as part of this PR would fail in 3.2 without marking the Array methods as `ruby2_keywords`.

geemus:
@DimitriosLisenko Oops, right. I guess my brain is still trying to get up to speed this morning. Could you rebase this on top of the addition of the 3.2 testing and reopen? Then we should hopefully be good to go. Sorry for the confusion.

DimitriosLisenko:
All good, re-opened here: https://github.com/fog/fog-core/pull/296

PR #296, opened 2023-12-22
Reapply "Delegate methods that accept keywords from Fog::Collection t…

DimitriosLisenko:
…o Array correctly for Ruby 2.x and Ruby 3.x (#292)" (#295)

This reverts commit 917bcb73f5a6c1c0f29fbf510345c5d8e4a1f490.

### DIFF
--- a/lib/fog/core/collection.rb
+++ b/lib/fog/core/collection.rb
@@ -19,6 +19,7 @@ module Fog
           end
           super
         end
+        ruby2_keywords(method) if respond_to?(:ruby2_keywords, true)
       EOS
     end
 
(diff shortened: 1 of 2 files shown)


geemus (approved):

geemus:
Thanks again. I imagine this deserves a release too?

DimitriosLisenko:
Agree, though I don't think it's necessarily urgent to release because most Array methods actually don't accept keywords, I think the only ones ATM are  `Array#shuffle`, `Array#sample`, and `Array#pack` (and maybe whatever else people have monkey patched Array with). Up to you, I imagine it can wait until after the holidays.

geemus:
Ah yeah, good point. Yep, my inclination was definitely to wait until the new year as I'm going to be otherwise occupied and would hate to accidentally push a bad release and be unavailable to address it.

geemus:
This should now be released in v2.4.0, thanks again!

DimitriosLisenko:
Thank you!

PR #300, opened 2024-06-15
Fix frozen string literal issue for ruby 3.4

chaadow:
@geemus Related to https://github.com/fog/fog-aws/pull/711

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,5 +12,21 @@ permissions:
   contents: read
 
 jobs:
-  Shared:
-    uses: fog/.github/.github/workflows/ci.yml@v1.4.0
+  test:
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        ruby-version: ['3.0', '3.1', '3.2', '3.3', 'head']
+    continue-on-error: ${{ matrix.ruby-version == 'head' }}
+
+    steps:
+    - uses: actions/checkout@v4
+    - name: Set up Ruby
+      uses: ruby/setup-ruby@v1
+      with:
+        ruby-version: ${{ matrix.ruby-version }}
+        bundler-cache: true # runs 'bundle install' and caches installed gems automatically
+    - name: Run tests
+      run: bundle exec rake RUBYOPT="--enable-frozen-string-literal"
--- a/spec/formatador_spec.rb
+++ b/spec/formatador_spec.rb
@@ -42,7 +42,7 @@ describe Fog::Formatador do
     end
 
     it "returns formatted representation" do
-      Fog::Formatador.format(@collection).must_equal @expected
+      _(Fog::Formatador.format(@collection)).must_equal @expected
     end
   end
 
@@ -65,7 +65,7 @@ describe Fog::Formatador do
     end
 
     it "returns formatted representation" do
-      Fog::Formatador.format(@collection).must_equal @expected
+      _(Fog::Formatador.format(@collection)).must_equal @expected
     end
   end
 
@@ -90,7 +90,7 @@ describe Fog::Formatador do
       EOS
 
       opts = { include_nested: false }
-      Fog::Formatador.format(@collection, opts).must_equal @expected
+      _(Fog::Formatador.format(@collection, opts)).must_equal @expected
     end
   end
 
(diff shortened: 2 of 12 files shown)


chaadow:
@geemus 
- I've decided to fix all frozen related issues in fog-core since I believe this is the gem that is most common to all `fog` dependencies.
- I also fixed some mini test warnings displayed by the test suite ( not all warnings)

- I had the same issue for my personal gem, but I think you need to enable CI again since it's disabled automatically by Github after 60 days of inactifvy https://github.com/fog/fog-core/actions/workflows/ci.yml

- I was wondering, since CI targets only ruby 3.0+, would you mind changing the target ruby version of `rubocop.yml`?
  - The main change is to add `# frozen_string_literal: true` to all files
    - Since this PR is related to this subject would you mind me adding it to all files, or a separate PR perhaps?

geemus:
Thanks for your continued efforts here.

- Makes sense to focus on fixing fog-core.
- Thanks for the minitest fixes.
- Yes, I didn't realize they disabled things after 60 days inactivity, I've reactivated it.
- I bumped the target in rubocop to 3.0 for now (we'll eventually want 3.3 I imagine, but I thought this would be a good next step).
- Feel free to add the string literal magic comment in this PR or a separate one, whatever is easier for you.

spec/formatador_spec.rb
```diff
@@ -42,7 +42,7 @@ def all
     end
 
     it "returns formatted representation" do
-      Fog::Formatador.format(@collection).must_equal @expected
+      _(Fog::Formatador.format(@collection)).must_equal @expected
```
geemus, line 45: I'm not familiar with this `_()` invocation, could you explain it to me?

.github/workflows/ci.yml
```diff
@@ -12,5 +12,21 @@ permissions:
   contents: read
 
 jobs:
-  Shared:
-    uses: fog/.github/.github/workflows/ci.yml@v1.4.0
```
geemus, line 16: Once this has settled, I think we'll want to go back to the shared action probably, so just noting that here so I don't forget.

geemus:
Looks good so far, just let me know when you need me to take another look.

voxik:
I was about submitting PR to drop the base64 and found it here (would be nice if it was independent PR IMHO). And I just wanted to point out that while fog-core does not use `base64` bit, other libraries from ecosystem might not load it by itself. However, those project CIs are likely broken by this anyway, so just heads up.

spec/formatador_spec.rb
ekohl: This is minitest and the old style is deprecated. https://minitest.rubystyle.guide/#global-expectations mentions this and https://docs.rubocop.org/rubocop-minitest/cops_minitest.html#minitestglobalexpectations can automatically replace them for you.

The alternative is to rewrite everything to `assert_equal()` & friends instead.

geemus: K, thanks. I hadn't bumped into that previously.

geemus:
@chaadow I'm forgetting now exactly where we left this (sorry). Is this where you wanted it to be, or was there still some work in progress? Just trying to figure out who should be waiting for whom since I can't quite remember. Thanks again for the help!

ekohl (commented):
Reviewing this change, it looks good to me. The only thing I can't answer is on the CI side.

The difference is that it passes `RUBYOPT="--enable-frozen-string-literal"` which the shared flow doesn't. There's a bit of a chicken-and-egg situation, though `head` is currently allowed to fail. So you could submit the change already. `head` is broken anyway (as can be seen in `fog-libvirt`) so it probably wouldn't get worse.

Other than that, it doesn't trigger CI. That is because it uses a schedule in the same job, so it gets deactivated if there's no change for 60 days.

Looks like it's enabled again so I'd at least advise rebasing it.

geemus:
@ekohl thanks for taking another look. I rebased and force pushed, which did in fact trigger CI, which passed. So I think we are (hopefully) in good shape now. I'll bring this in, just let me know if you'd like to work on any of the other libraries on related things, otherwise I suspect someone will get to it in time (and/or the upcoming Rubies will force the point early next year probably).

chaadow:
@geemus Hey, sorry i was on vacation, thanks for merging. I will take a look on the other related PRs that I've opened this weekend probably. Thanks again for your time!

PR #302, opened 2024-08-02
Exclude spec files from gem package

amatsuda:
Here's a patch that excludes files under spec directory from the gem package.

With this patch, the gem package size shrinks as follows:
```
before: 60416 bytes
after:  47104 bytes
```


geemus (approved):

geemus (approved):

geemus:
Thanks, I had done similar things on some of my gems and didn't realize I'd missed this one. I just did the same for fog-xml as well. Just let me know if you see any others and we'll get them fixed up also.

amatsuda:
@geemus Thanks! I'll do so!

